PR #4, opened 2022-02-23
'solution'

liuda-chuzdiuk:


alix-mak (approved):

app/tests/test_main.py
```diff
@@ -1,6 +1,6 @@
 import pytest
 
-from ..main import main, Genre, Actor
+from ..main import main, Genre, Actors
```
MasterpieceElbow, line 3: Typo in README.md was fixed, use `Actor` instead of `Actors`.

MasterpieceElbow (changes requested):

app/tests/test_main.py
liuda-chuzdiuk: fixed)

MasterpieceElbow (approved):

PR #23, opened 2022-06-12
Solution

rotsen18:


KatiukhaO (approved):

db/models.py
```diff
@@ -1 +1,10 @@
 from django.db import models
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255)
```

Magus6784 (changes requested):

db/models.py
Magus6784: This is literally words from task

Magus6784: Lol, where are they?!

Magus6784 (approved):
Change's on your side

PR #26, opened 2022-06-16
solution

DmytroSvirsa:


main.py
```diff
@@ -1,5 +1,30 @@
...
+
 
 def main():
-    pass
+
+    Genre.objects.create(name="Western")
+    Genre.objects.create(name="Action")
+    Genre.objects.create(name="Dramma")
+
+    Actor.objects.create(first_name="George", last_name="Klooney")
```
KatiukhaO, line 12: it`s correct, but, i think, will be better, create write in tables use loop, and write data from? some list or something else

KatiukhaO: here, broken DRY

KatiukhaO (approved):

db/models.py
```diff
@@ -1 +1,10 @@
 from django.db import models
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255)
```

Magus6784 (changes requested):

main.py
```diff
@@ -1,5 +1,34 @@
...
+              {"first_name": "Jaden", "last_name": "Smith"},
+              {"first_name": "Scarlett", "last_name": "Johansson"}]
+
+    for actor in actors:
+        Actor.objects.create(**actor)
+
+    Genre.objects.filter(
+        name="Dramma").update(name="Drama")
+    Actor.objects.filter(
+        last_name="Klooney").update(last_name="Clooney")
```
Magus6784, line 26: actor George Klooney, not just "Klooney"

Magus6784 (changes requested):

db/models.py
Magus6784: This is not actual now :(

Magus6784 (approved):

MasterpieceElbow (approved):

PR #27, opened 2022-06-16
Solution

KatiukhaO:


TPolina (approved):
you must push migrations to github as well. And don't delete `__init__.py`

KatiukhaO:
> you must push migrations to github as well. And don't delete `__init__.py`

oh, I understand, Can I add migration and __init__, now?

PR #30, opened 2022-06-16
Solution

RomaTsybulia:


main.py
```diff
@@ -1,5 +1,40 @@
 import init_django_orm  # noqa: F401
+from db.models import Genre, Actor
 
 
 def main():
-    pass
+    # create genres and actors
+    Genre.objects.create(name="Western")
```
KatiukhaO, line 7: it`s correct, but, i think, will be better, create write in tables use loop, and write data from, some list or something else

DRY

KatiukhaO (changes requested):

PR #31, opened 2022-06-16
Solution

Hectorovich:


db/models.py
```diff
@@ -1 +1,10 @@
 from django.db import models
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255)
```

Magus6784 (changes requested):
tests failed

Magus6784 (approved):

TPolina (approved):

PR #33, opened 2022-06-16
'Solution'

Aleksei-Isaev:


db/models.py
```diff
@@ -1 +1,10 @@
 from django.db import models
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255)
```

main.py
```diff
@@ -1,5 +1,31 @@
...
+    Genre.objects.create(name='Western')
+    Genre.objects.create(name='Action')
+    Genre.objects.create(name="Dramma")
+
+    Actor.objects.create(first_name="George", last_name="Klooney")
+    Actor.objects.create(first_name="Kianu", last_name="Reaves")
+    Actor.objects.create(first_name="Scarlett", last_name="Keegan")
+    Actor.objects.create(first_name="Will", last_name="Smith")
+    Actor.objects.create(first_name="Jaden", last_name="Smith")
+    Actor.objects.create(first_name="Scarlett", last_name="Johansson")
```
Magus6784, line 16: Why you are don't scared Python's God? Better create something like set or dict or list for data and make create with for

main.py
```diff
@@ -1,5 +1,31 @@
...
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+    Actor.objects.filter(last_name="Klooney").update(last_name="Clooney")
```
Magus6784, line 19: actor George Klooney, not just "Klooney"

Magus6784 (changes requested):

db/models.py
Aleksei-Isaev: Will be done as soon as possible)

main.py
Aleksei-Isaev: I just learn a new step and try to keep it simple.

main.py
```diff
@@ -1,5 +1,35 @@
...
+        first_name="Kianu"
+    ).update(first_name="Keanu", last_name="Reeves")
+
+    Genre.objects.filter(name="Action").delete()
+    Actor.objects.filter(first_name="Scarlett").delete()
+
+    return Actor.objects.filter(last_name="Smith").order_by("first_name")
+
+
+print(Genre.objects.all())
```
Abnormaltype, line 32: remove this code line

Abnormaltype (changes requested):

main.py
Aleksei-Isaev: I did it, but it doesn't work anyway. Somebody, could explain to me what I have to do? ))

PR #34, opened 2022-06-16
Katya's Solution

katevasilenko:


Chardona (approved):
Neat

TPolina (approved):
you should not push db to github. Change app/db.sqlite3 -> **db.sqlite3 in .gitignore

PR #37, opened 2022-06-16
Solution

Chardona:


main.py
```diff
@@ -1,5 +1,59 @@
...
+    )
+    Actor.objects.create(
+        first_name="Scarlett",
+        last_name="Johansson"
+    )
+    Genre.objects.filter(
+        name="Dramma"
+    ).update(name="Drama")
+    Actor.objects.filter(
+        last_name="Klooney"
```
Flashmobber, line 44: You need filtered by first_name too

Flashmobber (changes requested):

main.py
Chardona: Yes, indeed. My code works for this example, but in case of Clooney family, I would've gotten a mistake. Thanks.

Flashmobber (approved):

MasterpieceElbow (approved):

PR #62, opened 2022-07-12
Solution

MykolaIF:


db/models.py
```diff
@@ -1 +1,16 @@
 from django.db import models
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255)
+
+    def __repr__(self):
+        return f"<{type(self).__name__}: {self.name}>"
```
Abnormaltype, line 8: why do you need this?

db/models.py
```diff
@@ -1 +1,16 @@
...
+
+
+class Actor(models.Model):
+    first_name = models.CharField(max_length=255)
+    last_name = models.CharField(max_length=255)
+
+    def __repr__(self):
+        return f"<{type(self).__name__}: {self.first_name} {self.last_name}>"
```
Abnormaltype, line 16: the same question

Abnormaltype (changes requested):

db/models.py
MykolaIF: Because the following output format is specified in the example:
```
print(main())
# <QuerySet [<Actor: Jaden Smith>, <Actor: Will Smith>]>

print(Genre.objects.all())
# <QuerySet [<Genre: Western>, <Genre: Drama>]>

print(Actor.objects.all())
# <QuerySet [<Actor: George Clooney>, <Actor: Keanu Reeves>, <Actor: Will Smith>, <Actor: Jaden Smith>]>
```
Without __repr__, we have something like this:
```
<QuerySet [<Genre: Genre object (1)>, <Genre: Genre object (3)>]>
<QuerySet [<Actor: Actor object (1)>, <Actor: Actor object (2)>, <Actor: Actor object (4)>, <Actor: Actor object (5)>]>
```

Abnormaltype: it's not necessarily) But okay)

Abnormaltype (approved):
well done!

PR #99, opened 2022-10-04
Solution

AriohBelskij:


wQuelS (approved):
Everything is Ok.
But i'd put actors in tuple so you can iterate with 2 variables and assign them to first_name. last_name. So, you don't need to split. But anyway, let's see what mentor says

Y-Havryliv (approved):
Nice!

Y-Havryliv:
> Everything is Ok. But i'd put actors in tuple so you can iterate with 2 variables and assign them to first_name. last_name. So, you don't need to split. But anyway, let's see what mentor says

yep, sound good

PR #101, opened 2022-10-05
Solution

sbublinskyi:


main.py
```diff
@@ -1,5 +1,55 @@
...
+from db.models import Genre, Actor
 
 
 def main():
-    pass
+    create()
+    update()
+    delete()
+
+    return Actor.objects.filter(last_name="Smith").order_by("first_name")
```
mltdll, line 10: It's a shame that this `Retrieve` step here is not complicated enough for you to decide it's worth extracting a function. :)

mltdll (approved):

main.py
```diff
@@ -1,5 +1,55 @@
...
+def create():
+    genres = ["Western", "Action", "Dramma"]
+    actors = [
+        "George Klooney",
+        "Kianu Reaves",
+        "Scarlett Keegan",
+        "Will Smith",
+        "Jaden Smith",
+        "Scarlett Johansson"
+    ]
```
TPolina, line 22: why not create a list of tuples? You will not need to split each element later

TPolina (changes requested):

PR #106, opened 2022-10-07
Solution

AVyha:


main.py
```diff
@@ -1,7 +1,38 @@
...
+from db.models import Genre, Actor
 
 
 def main() -> QuerySet:
-    pass
+    Genre.objects.create(name="Western")
+    Genre.objects.create(name="Action")
+    Genre.objects.create(name="Dramma")
+
+    Actor.objects.create(first_name="George", last_name="Klooney")
```
mltdll, line 12: Why not use loops to create genres and actors?

main.py
```diff
@@ -1,7 +1,38 @@
...
+        first_name="Kianu", last_name="Reaves"
+    ).update(first_name="Keanu", last_name="Reeves")
+
+    Genre.objects.filter(name="Action").delete()
+
+    Actor.objects.filter(first_name="Scarlett").delete()
+
+    result = Actor.objects.filter(last_name="Smith").order_by("first_name")
+
+    return result
```
mltdll, line 34: I don't think that introducing the `result` variable improves readability in any way. I'd suggest to remove it.

main.py
```diff
@@ -1,7 +1,38 @@
...
+
+
+if __name__ == "__main__":
+    print(main())
```
mltdll, line 38: I don't think you should print it. I don't know if you should remove the print() here, but I would. The code would look cleaner this way.

mltdll (changes requested):

main.py
```diff
@@ -1,7 +1,34 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Genre, Actor
 
 
 def main() -> QuerySet:
-    pass
+    [Genre.objects.create(name=name) for name in ["Western",
```
mltdll, line 8: Huh. This style doesn't feel quite right to me. I would certainly use a traditional loop syntax here.

mltdll (changes requested):

mltdll (approved):

Nattalli (approved):
Good job!

PR #107, opened 2022-10-08
'Solution'

davidkrivko:


IvanRamyk (changes requested):
Tests should be passed. Add migrations to the pull request

IvanRamyk (approved):
Good job!

PR #108, opened 2022-10-09
Solution

olia-trofymchuk:


main.py
```diff
@@ -2,6 +2,50 @@
 
 from django.db.models import QuerySet
 
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
-    pass
+    # Create
+    Genre.objects.create(name="Western")
```
mltdll, line 10: I would strongly advise using loops here.

mltdll (changes requested):

main.py
olia-trofymchuk: Agree w u, thanks a lot. I forgot that we can use loops here.

mltdll (approved):

main.py
```diff
@@ -2,6 +2,54 @@
...
+        first_name="Scarlett").delete()
+
+    # Return
+    return Actor.objects.filter(
+        last_name="Smith"
+    ).order_by("first_name")
+
+
+if __name__ == "__main__":
+    main()
```
Nattalli, line 55: actually you don't need this fragment

Nattalli (changes requested):

main.py
olia-trofymchuk: okkaaaay)

Nattalli (approved):
Well done!!!

PR #109, opened 2022-10-09
Solution

IvanStored:


main.py
```diff
@@ -1,7 +1,39 @@
...
+              "Will Smith",
+              "Jaden Smith",
+              "Scarlett Johansson"]
+
+    for genre in genres:
+        Genre.objects.create(name=genre)
+
+    for actor in actors:
+        Actor.objects.create(
+            first_name=actor.split()[0],
```
wQuelS, line 24: you can loop over actors.split() with name, surname. So u don't need to split actors every time.
Or you can put actors in list of tuples E.g.  actors = [("George", "Klooney"),]
and loop with name, surname without split()

wQuelS (changes requested):

main.py
IvanStored: idk why i don`t do this at first solution, ty <3

wQuelS (approved):
Looks nice to me:)

TPolina (approved):
Good job!

PR #116, opened 2022-10-16
Develop

luxuriant777:


main.py
```diff
@@ -2,6 +2,36 @@
...
+
+    for genre in genres:
+        Genre.objects.create(name=genre)
+
+    actors = ["George Klooney",
+              "Kianu Reaves",
+              "Scarlett Keegan",
+              "Will Smith",
+              "Jaden Smith",
+              "Scarlett Johansson"]
```
TPolina, line 19: why not create a list of tuples? You will not need to use split later

main.py
```diff
@@ -2,6 +2,36 @@
...
+
+    for actor in actors:
+        Actor.objects.create(first_name=actor.split()[0],
+                             last_name=actor.split()[1])
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+    Actor.objects.filter(last_name="Klooney").update(last_name="Clooney")
+    Actor.objects.filter(first_name="Kianu", last_name="Reaves").update(
+        first_name="Keanu", last_name="Reeves")
```
TPolina, line 28: the close bracket should be from the next line:
```
...update(
    ...
)
```

TPolina (changes requested):

main.py
luxuriant777: please, check

main.py
luxuriant777: changed

TPolina (approved):
Good job!

PR #120, opened 2022-10-17
Solution

linaSam:


wQuelS (changes requested):
Just add migrations to your repo

main.py
```diff
@@ -1,7 +1,63 @@
...
+    )
+
+    Actor.objects.filter(
+        last_name="Klooney"
+    ).update(
+        last_name="Clooney"
+    )
+
+    Actor.objects.filter(
+        last_name="Reaves"
```
Y-Havryliv, line 43: what if there would be another actor with Reaves surname?

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Nice!

PR #123, opened 2022-10-17
Solution

VadimSabadyr:


main.py
```diff
@@ -1,7 +1,63 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Genre, Actor, Actress
 
 
 def main() -> QuerySet:
-    pass
+    Genre.objects.create(
```
mltdll, line 8: I think it's very reasonable to use loops in order to create genres and actors.

mltdll (changes requested):

main.py
```diff
@@ -5,39 +5,33 @@
...
+            name=genre
+        ),
+
+    actors = [
+        ["George", "Klooney"],
+        ["Kianu", "Reaves"],
+        ["Will", "Smith"],
+        ["Jaden", "Smith"]
+    ]
+    for actor in actors:
```
mltdll, line 20: You can also unpack values right here for more readability, but it's fine as is.

mltdll (approved):

main.py
```diff
@@ -1,7 +1,57 @@
...
+        Actor.objects.create(
+            first_name=actor[0],
+            last_name=actor[1]
+        ),
+
+    actresses = [
```
mltdll, line 26: What is not actually very good is this list and next iteration. Why don't yiu combine actors and actresses?

mltdll (changes requested):

main.py
VadimSabadyr: Так це різні таблички, навіщо їх змішувати?

mltdll (approved):
GJ! Sorry for missing some of the issues in the first iteration of the review.

Y-Havryliv (approved):
Nice!

PR #128, opened 2022-10-18
Solution

skrypka-dmytro:


main.py
```diff
@@ -2,6 +2,53 @@
...
+    actors = [
+        ("George", "Klooney"), ("Kianu", "Reaves"), ("Scarlett", "Keegan"),
+        ("Will", "Smith"), ("Jaden", "Smith"), ("Scarlett", "Johansson")
+    ]
+
+    genres = ["Western", "Action", "Dramma"]
+
+    for names in actors:
+        Actor.objects.create(
+            first_name=names[0],
```
wQuelS, line 18: u have list of tuples, so why don't u iterate with name, surname. So you don't have to use indexes. Read about unpacking or just read cheklist for this task https://github.com/mate-academy/py-genres-and-actors/blob/master/checklist.md

main.py
```diff
@@ -2,6 +2,53 @@
...
+        Genre.objects.create(
+            name=genre
+        )
+
+    Genre.objects.filter(
+        name="Dramma",
+    ).update(name="Drama")
+
+    Actor.objects.filter(
+        last_name="Klooney"
```
wQuelS, line 32: Task says that we have to update surname for George Klooney. If there was another guy with the same surname, but with name that differs - personal info should remain the same.
Just like u've done it with Keanu Reeves

wQuelS (changes requested):
Check the comments below.
Also u can try to use black for better code formatting. It can be simply installed via pip
https://pypi.org/project/black/

Y-Havryliv (approved):
Nice!

PR #133, opened 2022-10-18
Solution)

Gr1zzZzzly:


main.py
```diff
@@ -3,5 +3,36 @@
...
+        "Jaden Smith",
+        "Scarlett Johansson"
+    ]
+
+    for genre in genres:
+        Genre.objects.create(name=genre)
+
+    for actor in actors:
+        Actor.objects.create(
+            first_name=actor.split()[0],
```
wQuelS, line 26: You don't need to split() every time. Just use unpacking.
Here some tips https://github.com/mate-academy/py-genres-and-actors/blob/master/checklist.md

main.py
```diff
@@ -3,5 +3,36 @@
...
+            last_name=actor.split()[1]
+        )
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+    Actor.objects.filter(last_name="Klooney").update(last_name="Clooney")
```
wQuelS, line 31: If there is another guy with same surname but his name differs - his personal info should not be changed

wQuelS (changes requested):

wQuelS (approved):

TPolina (approved):
Good job!

PR #142, opened 2022-10-19
Solution

Mowka7:


main.py
```diff
@@ -1,7 +1,43 @@
...
+    actors = [
+        ["George", "Klooney"],
+        ["Kianu", "Reaves"],
+        ["Scarlett", "Keegan"],
+        ["Will", "Smith"],
+        ["Jaden", "Smith"],
+        ["Scarlett", "Johansson"]
+    ]
+
+    for actor in actors:
```
mltdll, line 22: You can unpack lists in place, if you want. For example,
```python
for value1, value2 in list_of_pairs:

mltdll (approved):

main.py
mltdll: You can sure use a dictionary if you want to, but I don't think that's necessary. I don't even think that would be an improvement.

Mowka7: Ok, I'll leave it like this, then

TPolina: you don't need a dictionary to unpack tuples in the cycle)

TPolina (approved):
Good job!

PR #143, opened 2022-10-19
Solution

fsocie7y:


main.py
```diff
@@ -1,7 +1,51 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Genre, Actor
 
 
 def main() -> QuerySet:
-    pass
+    # CREATE
```
LiudmylaKulinchenko, line 8: You don't need these comments

main.py
```diff
@@ -1,7 +1,51 @@
...
+        ("Scarlett", "Johansson")
+    ]
+    for first_name, last_name in actors:
+        Actor.objects.create(
+            first_name=first_name,
+            last_name=last_name
+        )
+
+    # UPDATE
+    Genre.objects.filter(name="Dramma").update(name="Drama")
```
LiudmylaKulinchenko, line 28: I'm not sure about using here two different ways to update

LiudmylaKulinchenko (approved):
Looks great for me 👍

main.py
```diff
@@ -1,7 +1,49 @@
...
+
+    genre = Genre.objects.get(name="Dramma")
+    genre.name = "Drama"
+    genre.save()
+
+    actor1 = Actor.objects.get(
+        first_name="George", last_name="Klooney"
+    )
+    actor1.last_name = "Clooney"
+    actor1.save()
```
TPolina, line 34: you could do it easier with `filter` and `update` but ok

TPolina: but it's strange that you use two different approaches in one code

TPolina (approved):
Consider my comment

PR #145, opened 2022-10-19
Solution

LiudmylaKulinchenko:


fsocie7y (approved):
Greate, i dont't see something wrong)

db/models.py
```diff
@@ -1 +1,10 @@
-from django.db import models
+from django.db import models  # nota:F401
```
TPolina, line 1: you don't need this comment here

TPolina (approved):
Good job!

PR #147, opened 2022-10-19
solution

ShevchukVitalii:


main.py
```diff
@@ -1,7 +1,33 @@
...
+        Actor.objects.create(
+            first_name=firstname,
+            last_name=lastname
+        )
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+    Actor.objects.filter(last_name="Klooney").update(last_name="Clooney")
+    Actor.objects.filter(
+        first_name="Kianu",
+        last_name="Reaves").update(first_name="Keanu", last_name="Reeves")
```
HStakhiv, line 28: It is better to do the transfer like this:
```
Actor.objects.filter(
    ....
    ....
).update(...)
```

main.py
```diff
@@ -1,7 +1,33 @@
...
+              ("Jaden", "Smith"),
+              ("Scarlett", "Johansson")]
+    for firstname, lastname in actors:
+        Actor.objects.create(
+            first_name=firstname,
+            last_name=lastname
+        )
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+    Actor.objects.filter(last_name="Klooney").update(last_name="Clooney")
```
HStakhiv, line 25: We need to change the last name only for `George Klooney`

HStakhiv (changes requested):

HStakhiv (approved):

Nattalli (approved):
Good!

PR #148, opened 2022-10-19
Solution

dmytro-petliovanyi:


main.py
```diff
@@ -2,6 +2,28 @@
 
 from django.db.models import QuerySet
 
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
-    pass
+    Genre.objects.create(name="Western")
```
Vasyl-Poremchuk, line 9: Your code is repeating itself. Imagine that you have a file and you need to write all the information in the database. You will have to perform the same operation many times, which is not optimal.  I recommend putting all the inputs into a tuple  `("Western", "Action", "Dramma")` and then iterating over them, and then adding the values to the corresponding table columns.

main.py
```diff
@@ -2,6 +2,28 @@
...
+    Genre.objects.create(name="Action")
+    Genre.objects.create(name="Dramma")
+
+    Actor.objects.create(first_name="George", last_name="Klooney")
```
Vasyl-Poremchuk, line 13: The same here. `((first_name, last_name), (..., ...), ...)`

Vasyl-Poremchuk (changes requested):

PR #150, opened 2022-10-19
Solution

FranchukS:


OvDima1 (approved):
Looks good!

main.py
```diff
@@ -2,6 +2,49 @@
...
-    pass
+    genres = ("Western", "Action", "Dramma")
+    actors = (
+        "George Klooney",
+        "Kianu Reaves",
+        "Scarlett Keegan",
+        "Will Smith",
+        "Jaden Smith",
+        "Scarlett Johansson"
+    )
```
TPolina, line 17: why not create a list of tuples? you will not need to use split later

main.py
```diff
@@ -2,6 +2,49 @@
...
+        last_name="Reeves"
+    )
+
+    Genre.objects.filter(name="Action").delete()
+
+    Actor.objects.filter(
+        first_name="Scarlett"
+    ).delete()
+
+    return Actor.objects.filter(last_name="Smith").all().order_by("first_name")
```
TPolina, line 50: `all` in this query is redundant

TPolina (changes requested):

main.py
FranchukS: Emm... cause I confused good and bad practice examples in checklist. And it was late so I already didn't think by myself😅

TPolina (approved):
Good job!

PR #235, opened 2023-02-15
Solution

Devrro:
Hey there


db/models.py
```diff
@@ -1 +1,11 @@
 from django.db import models
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255)
+
+
+class Actor(models.Model):
+
```
thorium-zero, line 9: a bit inconsistent styling in comparison to the previous model, you should not use an empty line after the class signature, just as you did with the previous model

main.py
```diff
@@ -1,7 +1,88 @@
...
+    for genre in list_of_genres:
+        Genre.objects.create(
+            name=f"{genre}"
+        )
+
+    for actor in list_of_actors:
+        Actor.objects.create(
+            first_name=actor["first_name"],
+            last_name=actor["last_name"]
+        )
```
thorium-zero, line 50: You could use bulk_create method to... umm... create objects in bulk :) That will increase your query performance greatly

main.py
```diff
@@ -1,7 +1,88 @@
...
+    {
+        "first_name": "Jaden",
+        "last_name": "Smith"
+    },
+    {
+        "first_name": "Scarlett",
+        "last_name": "Johansson"
+    },
+
+]
```
thorium-zero, line 36: You could probably use lists of tuples for the source of actor, that will allow you to write somewhat less verbose code in this case, although if you expect to use another types of sources in the future, like json objects for example, this approach could be better

main.py
```diff
@@ -1,7 +1,88 @@
...
+        first_name="Scarlett"
+    ).delete()
+    # Return QuerySet of actors with `last_name`
+    # "Smith" and ordered by `first_name`
+    query = Actor.objects.filter(
+        last_name="Smith"
+    ).order_by(
+        "first_name"
+    )
+    return query
```
thorium-zero, line 88: There's no need to use an intermediate reference to the query if the only use for that is to return its value

thorium-zero (approved):
There are some subjective issues here. All in all, the solution is acceptable

thorium-zero (approved):

thorium-zero (approved):

YevheniiIovchu (approved):

erikagayan (approved):

AntonShpakovych (approved):

Tarasidze:
Nice)

Tarasidze (approved):
Nice)

OlhaTryhub (approved):
🤩

ant-komarov (approved):

imelnyk007 (approved):

IhorVoskoboinikov (approved):
GJ. I think that you must use list_of_genres and list_of_actors as constant with uppercase like:
LIST_OF_GENRES and LIST_OF_ACTORS.

yakhlamova (approved):

YKonoplyov (approved):

Kirontiko (approved):

DmytroTsehelnyk (approved):

meforoma (approved):
[technical approve]

PR #254, opened 2023-03-01
Solution

yura-hudzovskyi:


WDemidenko (approved):
Well done!

main.py
```diff
@@ -1,7 +1,38 @@
...
+    actors = [
+        ("George", "Klooney"),
+        ("Kianu", "Reaves"),
+        ("Scarlett", "Keegan"),
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson")
+    ]
+
+    [
```
Y-Havryliv, line 18: right now you are creating a list that you won't use anywhere, so in this case, it's better to use just the for-loop, not list comprehension

Y-Havryliv (changes requested):

usrofgh (approved):

wQuelS (changes requested):
Good

wQuelS (approved):

PR #274, opened 2023-03-07
Solution

ZhAlexR:


inrock1 (approved):
i think you must push changed main.py
now in your repository Data list in two places.
The other looks good.

main.py
```diff
@@ -2,6 +2,28 @@
...
 
 def main() -> QuerySet:
-    pass
+    genres, actors = DATA
+
+    for genre in genres:
+        Genre.objects.create(name=genre)
+
+    for actor in actors:
+        first_name, last_name = actor.split()
```
Serhii-Leonenko, line 16: It's better to store first name and last name in the tuples according to the checklist because the split method takes an extra time

Serhii-Leonenko (changes requested):

main.py
```diff
@@ -2,6 +2,27 @@
...
+
 
 def main() -> QuerySet:
-    pass
+
+    for genre in GENRES:
+        Genre.objects.create(name=genre)
+
+    for actor in ACTORS:
+        first_name, last_name = actor
```
YuriiKindrat, line 15: u can unpack actor variable in loop inicialization 
like: `first_name, last_name in Acotrs`

main.py
```diff
@@ -2,6 +2,27 @@
...
+        Actor.objects.create(first_name=first_name, last_name=last_name)
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+    Actor.objects.filter(last_name="Klooney").update(last_name="Clooney")
+    Actor.objects.filter(first_name="Kianu").update(
+        first_name="Keanu",
+        last_name="Reeves"
```
YuriiKindrat, line 22: what if db will have other Klooneys ?
image
u should make update only for George, the same is for Kianu

YuriiKindrat (changes requested):

wQuelS (approved):

PR #277, opened 2023-03-07
Solutions ns about warnings, seems system issues

NEkropo1:


oksanaaam (approved):

main.py
```diff
@@ -2,6 +2,55 @@
...
+    Actor.objects.filter(first_name="George",
+                         last_name="Klooney").update(
+        last_name="Clooney"
+    )
+
+    Actor.objects.filter(first_name="Kianu",
+                         last_name="Reaves").update(
+        first_name="Keanu",
+        last_name="Reeves"
+    )
```
Serhii-Leonenko, line 43: it's better to use the same style for breaking lines

Serhii-Leonenko (approved):

main.py
NEkropo1: I just don't like look of separated `.update()` so didn't touch it, but it's still more readable after same breaking like styling, so I refactored that )

PR #289, opened 2023-03-08
solution

Yulia-Mala:


oksana0410 (approved):

main.py
```diff
@@ -2,6 +2,41 @@
...
+
+    for genre in genres:
+        Genre.objects.create(name=genre)
+
+    for name, surname in actors:
+        Actor.objects.create(first_name=name, last_name=surname)
+
+    drama = Genre.objects.get(name="Dramma")
+    drama.name = "Drama"
+    drama.save()
```
Serhii-Leonenko, line 25: Try to do the same without a redundant variable in one line with the method `update()`. The same in other similar cases

main.py
```diff
@@ -2,6 +2,41 @@
...
+    keanu = Actor.objects.get(first_name="Kianu", last_name="Reaves")
+    keanu.first_name = "Keanu"
+    keanu.last_name = "Reeves"
+    keanu.save()
+
+    Genre.objects.get(name="Action").delete()
+
+    scarlett_ls = Actor.objects.filter(first_name="Scarlett")
+    for person in scarlett_ls:
+        person.delete()
```
Serhii-Leonenko, line 40: for loop is redundant here

Serhii-Leonenko (changes requested):

main.py
Yulia-Mala: done

main.py
Yulia-Mala: there is a line in README "delete all actresses with the first_name "Scarlett". So it implies that it could be more than one instance. That's why loop is here

Yulia-Mala: now I see, delete also works with QuerySet not just with one instance. Thanks)

Serhii-Leonenko (approved):
Good job!

PR #294, opened 2023-03-08
'Solution'

alexlega:


myroslavatrschk (changes requested):
Remove file "db.sqlite3" from your commit.

YuriiKindrat (changes requested):

main.py
```diff
@@ -1,7 +1,37 @@
 import init_django_orm  # noqa: F401
 
+from db.models import Actor, Genre
 from django.db.models import QuerySet
```
Serhii-Leonenko, line 4: Fix imports according to PEP8

https://peps.python.org/pep-0008/#imports

main.py
```diff
@@ -1,7 +1,37 @@
...
+    ]
+
+    for first_name, last_name in actors:
+        Actor.objects.create(
+            first_name=first_name,
+            last_name=last_name
+        )
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+    Actor.objects.filter(last_name="Klooney").update(last_name="Clooney")
```
Serhii-Leonenko, line 29: It's working in this case, but it's better to add the first name to the filter according to the task 

image

Serhii-Leonenko (changes requested):

wQuelS (approved):

PR #298, opened 2023-03-08
Solution DonchykD

DmytryTong:


main.py
```diff
@@ -2,6 +2,52 @@
...
+            last_name=last_name_,
+        )
+
+    Genre.objects.filter(
+        name="Dramma",
+    ).update(name="Drama")
+
+    Actor.objects.filter(
+        last_name="Klooney",
+    ).update(last_name="Clooney")
```
AlessaQ, line 34: The condition says you have to change George Klooney's surname, not all Klooney surnames.

main.py
```diff
@@ -2,6 +2,52 @@
...
+
+    Actor.objects.filter(
+        first_name="Scarlett",
+    ).delete()
+
+    result = Actor.objects.filter(
+        last_name="Smith"
+    ).order_by("first_name")
+
+    return result
```
AlessaQ, line 53: It is not necessary to create a new variable, it makes sense to return the list directly.

main.py
```diff
@@ -2,6 +2,52 @@
...
 
 def main() -> QuerySet:
-    pass
+    genre_names = ["Western", "Action", "Dramma"]
+    actors = [("George", "Klooney"),
+              ("Kianu", "Reaves"),
+              ("Scarlett", "Keegan"),
+              ("Will", "Smith"),
+              ("Jaden", "Smith"),
+              ("Scarlett", "Johansson")]
```
AlessaQ, line 15: Not the best way to transfer lines.

AlessaQ (changes requested):

main.py
DmytryTong: In checklist - recomended list of tuples.

main.py
DmytryTong: Good tought - fixed.

main.py
DmytryTong: Yup, good touhgt to) Fixed.

YuriiKindrat (approved):

PR #303, opened 2023-03-08
Solution

myroslavatrschk:


main.py
```diff
@@ -2,6 +2,35 @@
...
+
+    for genre in genres:
+        Genre.objects.create(name=genre)
+
+    for first_name, last_name in actors:
+        Actor.objects.create(first_name=first_name, last_name=last_name)
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+
+    Actor.objects.filter(first_name="George").update(last_name="Clooney")
```
Serhii-Leonenko, line 27: It's working, but it's better to add the last name to the filter for correct filtering

main.py
```diff
@@ -2,6 +2,35 @@
...
+    Actor.objects.filter(first_name="Kianu").update(
+        first_name="Keanu", last_name="Reeves"
+    )
```
Serhii-Leonenko, line 30: The same is here

Serhii-Leonenko (changes requested):

main.py
```diff
@@ -2,6 +2,35 @@
...
+
+    for genre in genres:
+        Genre.objects.create(name=genre)
+
+    for first_name, last_name in actors:
+        Actor.objects.create(first_name=first_name, last_name=last_name)
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+
+    Actor.objects.filter(last_name="Klooney").update(last_name="Clooney")
```
Serhii-Leonenko, line 27: I meant that you should filter by the first name and the last name according to the logic you should change the last name for the particular actor

Serhii-Leonenko (changes requested):

wQuelS (approved):
Good

PR #314, opened 2023-03-13
Solution

AnnaBoichenko:


main.py
```diff
@@ -2,6 +2,46 @@
...
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
-    pass
+    genres = ["Western", "Action", "Dramma"]
+    actors = [("George", "Klooney"), ("Kianu", "Reaves"),
+              ("Scarlett", "Keegan"), ("Will", "Smith"),
+              ("Jaden", "Smith"), ("Scarlett", "Johansson")
+              ]
```
YuriiKindrat, line 13: fix indentation here

main.py
```diff
@@ -2,6 +2,46 @@
...
+            last_name=actor[1]
+        )
+
+    Genre.objects.filter(
+        name="Dramma"
+    ).update(name="Drama")
+
+    Actor.objects.filter(
+        last_name="Klooney"
+    ).update(last_name="Clooney")
```
YuriiKindrat, line 32: What if it will be another Klooney in db? u have to update George according to the task

YuriiKindrat (changes requested):

PR #318, opened 2023-03-17
Solution

lotrik84:


ArsenPidhoretskyi (changes requested):
Migration file does not exist

ArsenPidhoretskyi (approved):
Well done!

PR #327, opened 2023-03-29
Develop

yuraua5:


main.py
```diff
@@ -2,6 +2,30 @@
 
 from django.db.models import QuerySet
 
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
-    pass
+    genres = ["Western", "Action", "Dramma"]
+    [Genre.objects.create(name=genre) for genre in genres]
```
Grayder0152, line 10: Do not use list compl if you will not use it in future

Grayder0152 (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #334, opened 2023-04-05
Solution

Bloodviel:


db/models.py
```diff
@@ -1 +1,10 @@
 from django.db import models
+
+
+class Genre(models.Model):
```
vasylhnatiuk, line 4: please add str method for models

vasylhnatiuk (changes requested):

Grayder0152 (approved):

PR #398, opened 2023-06-27
solution

mariiahorbova:


main.py
```diff
@@ -2,6 +2,37 @@
...
+        ("Scarlett", "Keegan"),
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson")
+    ]
+    for item_first, item_second in actors:
+        Actor.objects.create(
+            first_name=item_first,
+            last_name=item_second
+        )
```
Kirontiko, line 25: I think u can change it to generator, but it's up to you

Kirontiko: Also for best readability you can change item_first, item_second to first_name, last_name

main.py
```diff
@@ -2,6 +2,37 @@
...
 from django.db.models import QuerySet
 
+from db.models import Actor, Genre
+
 
 def main() -> QuerySet:
-    pass
+    genres = ["Western", "Action", "Dramma"]
+    for genre in genres:
+        Genre.objects.create(name=genre)
```
Kirontiko, line 11: U can transform here to comprehension also

Kirontiko (approved):

RedCatHot (approved):

main.py
YuriiKindrat: @Kirontiko is right, it is better to name variables as first_name, last_name for better readability

PR #400, opened 2023-06-28
Solution

DmytroTsehelnyk:


main.py
```diff
@@ -2,6 +2,54 @@
...
+        Genre.objects.create(name=genre_name)
+
+    names = [
+        ("George", "Klooney"),
+        ("Kianu", "Reaves"),
+        ("Scarlett", "Keegan"),
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson"),
+    ]
```
IhorVoskoboinikov, line 22: I think that "genres" and "names" it is a constant. Use it in global namespace. I think it will be better)))

DmytroTsehelnyk: It's unnecessary to use it in the global namespace, 
because these lists affect only the main function, so it's more logical to keep them here

main.py
```diff
@@ -2,6 +2,54 @@
 
 from django.db.models import QuerySet
 
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
-    pass
+
+    # Create
```
LiudmylaKulinchenko, line 10: remove comments

main.py
```diff
@@ -2,6 +2,54 @@
...
+    genres = ["Western", "Action", "Dramma"]
+    for genre_name in genres:
+        Genre.objects.create(name=genre_name)
+
+    names = [
```
LiudmylaKulinchenko, line 15: isn't it better to rename to `actors`? What do you think?
Probably not all of names will be actors - that what developer could think when look at this
Try to clarily

LiudmylaKulinchenko, line 15: assigning variable names should be at the start of function

main.py
```diff
@@ -2,6 +2,54 @@
...
+    ).delete()
+
+    Actor.objects.filter(
+        first_name="Scarlett"
+    ).delete()
+
+    # Return
+    return Actor.objects.filter(
+        last_name="Smith"
+    ).order_by("first_name").all()
```
LiudmylaKulinchenko, line 55: no need to use `all` after `order_by`

LiudmylaKulinchenko (changes requested):

main.py
```diff
@@ -2,6 +2,49 @@
...
+    Actor.objects.filter(
+        last_name="Klooney"
+    ).update(last_name="Clooney")
+
+    Actor.objects.filter(
+        first_name="Kianu"
+    ).update(first_name="Keanu",
+             last_name="Reeves")
+
+    Genre.objects.filter(
```
LiudmylaKulinchenko, line 40: You don't need to split in lines everywhere

LiudmylaKulinchenko (approved):
Good job! 😋😸

PR #401, opened 2023-06-28
Solution #1

erikagayan:


Tarasidze (changes requested):
You need to add migrations to your pull request. You may use: "git add PATH_TO _YOUR_FILE" command or change .gitignore file.
Everything else is ok!

LiudmylaKulinchenko (changes requested):
tests didn't pass

main.py
```diff
@@ -1,7 +1,57 @@
...
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson"),
+    ]
+
+    for actor in actors:
+        Actor.objects.create(
+            first_name=actor[0],
+            last_name=actor[1]
+        )
```
Oleg-s-s, line 32: move variables to top and then apply some operations. That will make your code easier to understand

Oleg-s-s (changes requested):
Taras mentioned that you need migrations on a pr. Thats true but holding db itself on git is a crucial mistake. remove db from your pr

Oleg-s-s (changes requested):
You still left your db in pr, it should be removed

PR #404, opened 2023-06-28
Genres and Actions: Django ORM (first commit)

IhorVoskoboinikov:


diana-shyrokikh (approved):

main.py
```diff
@@ -1,7 +1,45 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Actor, Genre
+
+GENRE_NAMES = ["Western", "Action", "Dramma"]
+ACTORS = [
```
LiudmylaKulinchenko, line 7: Grobal variables allow data to be shared and used across multiple functions or modules. You don't need it here

LiudmylaKulinchenko: `actors` could be list of tuples, so you'd iterate by them as in the checklist example

main.py
```diff
@@ -1,7 +1,45 @@
...
+
+    Actor.objects.filter(
+        first_name="Kianu", last_name="Reaves"
+    ).update(first_name="Keanu", last_name="Reeves")
+
+    Genre.objects.filter(name="Action").delete()
+
+    Actor.objects.filter(first_name="Scarlett").delete()
+
+    actors_with_last_name_smith = Actor.objects.filter(
```
LiudmylaKulinchenko, line 41: you could consider shortening it to something like "smith_actors" or "smiths" if you prefer a shorter name while still maintaining clarity

LiudmylaKulinchenko (changes requested):

Oleg-s-s (approved):
:+1:

PR #411, opened 2023-06-29
'Solution'

yakhlamova:


main.py
```diff
@@ -1,7 +1,38 @@
...
+        first_name="Kianu", last_name="Reaves"
+    ).update(first_name="Keanu", last_name="Reeves")
+
+    Genre.objects.filter(name="Action").delete()
+    Actor.objects.filter(first_name="Scarlett").delete()
+
+    filtered_smith = Actor.objects.filter(
+        last_name="Smith"
+    ).order_by("first_name")
+    return filtered_smith
```
Andriy-Sydorenko, line 38: You don't necessarily need to create the variable `filtered_smith` in my opinion, just return the query set without it.

Andriy-Sydorenko (approved):
Good Job!)

main.py
LiudmylaKulinchenko: I agree, you could skip it. What do you think?

LiudmylaKulinchenko: Better to add an empty line before return statement

LiudmylaKulinchenko (approved):
Well done! 🌺👏

main.py
yakhlamova: Understood, I will remember it 🌺

PR #416, opened 2023-06-29
Implemented Genre, Actor models

Altahoma:


DmytroTsehelnyk (approved):
Interesting solution with 'bulk_create'. GJ!

PR #417, opened 2023-06-29
Solution

Po-SvinSkipodla:


main.py
```diff
@@ -1,7 +1,32 @@
...
+    items = [
+        ("George", "Klooney"),
+        ("Kianu", "Reaves"),
+        ("Scarlett", "Keegan"),
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson")
+    ]
+    for first, last in items:
+        Actor.objects.create(first_name=first, last_name=last)
```
DmytroTsehelnyk, line 20: In this case, you can use "name" and "surname" name instead of "first" and "last".

main.py
```diff
@@ -1,7 +1,32 @@
...
 
 from django.db.models import QuerySet
 
 
 def main() -> QuerySet:
-    pass
+    items = ["Western", "Action", "Dramma"]
+    for item in items:
+        Genre.objects.create(name=item)
+    items = [
```
DmytroTsehelnyk, line 11: Same problem here. Use "actor_names" or "names" instead of "items"

main.py
```diff
@@ -1,7 +1,32 @@
 import init_django_orm  # noqa: F401
+from db.models import Genre, Actor
 
 from django.db.models import QuerySet
 
 
 def main() -> QuerySet:
-    pass
+    items = ["Western", "Action", "Dramma"]
```
DmytroTsehelnyk, line 8: It will be more logical to use "movie_names" or "movie_titles" instead of "items"

DmytroTsehelnyk (changes requested):

main.py
Po-SvinSkipodla: I think my variant is better, just because i don't need to create 3 variables. Just create one and rewrite it. Memory economy)

DmytroTsehelnyk (approved):

main.py
```diff
@@ -1,7 +1,32 @@
...
+    Actor.objects.filter(last_name="Klooney").update(last_name="Clooney")
+    Actor.objects.filter(
+        first_name="Kianu",
+        last_name="Reaves"
+    ).update(
+        first_name="Keanu",
+        last_name="Reeves"
+    )
+    Genre.objects.filter(name="Action").delete()
+    Actor.objects.filter(first_name="Scarlett").delete()
```
YuriiKindrat, line 31: add some blank lines here, for better code style
also, what if we will have another actor with the last name Klooney?
according to the task, u should cahnge a last name for George 
image

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Well done! 😋

PR #450, opened 2023-07-28
Implemented function main

lubart2517:


db/models.py
```diff
@@ -1 +1,10 @@
 from django.db import models
+
+
+class Genre(models.Model):
```
Polyakiv-Andrey, line 4: Add str method to all models

db/models.py
```diff
@@ -1 +1,10 @@
...
+    name = models.CharField(max_length=255)
+
+
+class Actor(models.Model):
```
Polyakiv-Andrey, line 8: add str method to all models

Polyakiv-Andrey (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #452, opened 2023-08-02
Solved

Dimo-n:


db/models.py
```diff
@@ -1 +1,10 @@
 from django.db import models
+
+
+class Genre(models.Model):
```
Polyakiv-Andrey, line 4: add str method

db/models.py
```diff
@@ -1 +1,10 @@
...
+    name = models.CharField(max_length=255)
+
+
+class Actor(models.Model):
```

Polyakiv-Andrey (changes requested):

main.py
```diff
@@ -2,6 +2,31 @@
 
 from django.db.models import QuerySet
 
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
```
Dimosphen1, line 8: You could specify that you expect to return the queryset of `Actor` model.

PR #454, opened 2023-08-04
solution

sbilikepy:


sbilikepy:
Regarding point 2 of the checklist from the README:

Put values in the list of tuples for better iterating.
_Good example (list of tuples):_
```
for item_first, item_second in items:
    Model.objects.create(
        first=item_first,
        second=item_second
    )
```
I decided to keep the list of strings since it's more likely that the data will come in string format rather than tuples. Is this correct, or do I need to make changes?

db/models.py
```diff
@@ -1 +1,16 @@
 from django.db import models
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255)
+
+    def __repr__(self) -> str:
+        return f"<{__class__.__name__}: {self.name}>"
```
ArsenPidhoretskyi, line 8:
```suggestion
        return f"<{self.__class__.__name__}: {self.name}>"
```

db/models.py
```diff
@@ -1 +1,16 @@
...
+
+
+class Actor(models.Model):
+    first_name = models.CharField(max_length=255)
+    last_name = models.CharField(max_length=255)
+
+    def __repr__(self) -> str:
+        return f"<{__class__.__name__}: {self.first_name} {self.last_name}>"
```
ArsenPidhoretskyi, line 16:
```suggestion
        return f"<{self.__class__.__name__}: {self.first_name} {self.last_name}>"
```

ArsenPidhoretskyi (approved):
Good job!

PR #456, opened 2023-08-08
'Solution'

olmatiash:


db/migrations/0001_initial.py
```diff
@@ -0,0 +1,29 @@
...
+class Migration(migrations.Migration):
+
+    initial = True
+
+    dependencies = [
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='Actor',
```
vasylhnatiuk, line 15: use `'` instead of `"`, use Black

vasylhnatiuk (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #462, opened 2023-08-18
'Solution'

niksgawrilenko:


db/models.py
```diff
@@ -1 +1,10 @@
 from django.db import models
+
+
+class Genre(models.Model):
```
Polyakiv-Andrey, line 4: add str method

db/models.py
```diff
@@ -1 +1,10 @@
...
+    name = models.CharField(max_length=255)
+
+
+class Actor(models.Model):
```

Polyakiv-Andrey (changes requested):

db/models.py
niksgawrilenko: done

db/models.py
niksgawrilenko: done

PR #483, opened 2023-09-07
Implemented correct data-table editing

ORENYT:
- Implemented the "put_values" function to simplify creating data
- Data written in tuples
- Update and delete made without extra code
- Models are changed to represent correct string data
- Checklist checked


denys-source (approved):
I will approve, but mentors will decide if `put_values` method necessary in this task

PR #486, opened 2023-09-07
Solution

CaCtUs1k:


main.py
```diff
@@ -2,6 +2,35 @@
 
 from django.db.models import QuerySet
 
+from db.models import Actor, Genre
+
 
 def main() -> QuerySet:
-    pass
+    Genre.objects.all().delete()
```
MarianKovalyshyn, line 9: I don't understand why you try to delete data from tables if your tables are empty.

main.py
```diff
@@ -2,6 +2,35 @@
...
+    Actor.objects.all().delete()
+    genres = ["Western", "Action", "Dramma"]
+    for genre in genres:
+        Genre.objects.create(name=genre)
+    actors = [["George", "Klooney"],
```
MarianKovalyshyn, line 14: In checklist it was said to use tuples inside of a list to use it in for loop more clearly, like that:
```
for item_first, item_second in items:
       Model.objects.create(
           first=item_first,
           second=item_second
       )
```

MarianKovalyshyn (approved):

PR #508, opened 2023-09-18
Solution

AlexPelevin99:


viktoria-rybenchuk (changes requested):

db/models.py
```diff
@@ -1 +1,19 @@
...
+
+
+class BaseModel:
+    @classmethod
+    def create(cls, items: List[Dict]) -> None:
+        items = [cls(**item) for item in items]
+        cls.objects.bulk_create(items)
+
+
+class Genre(BaseModel, models.Model):
```
Polyakiv-Andrey, line 13: add str method

db/models.py
```diff
@@ -1 +1,19 @@
...
+    name = models.CharField(max_length=255)
+
+
+class Actor(BaseModel, models.Model):
```
Polyakiv-Andrey, line 17: add str method

Polyakiv-Andrey (changes requested):

main.py
```diff
@@ -1,7 +1,44 @@
...
+    ).update(last_name="Clooney")
+    Actor.objects.filter(
+        first_name="Kianu",
+        last_name="Reaves"
+    ).update(first_name="Keanu", last_name="Reeves")
+
+    Genre.objects.filter(name="Action").delete()
+    Actor.objects.filter(first_name="Scarlett").delete()
+
+    return Actor.objects.filter(last_name="Smith").order_by("first_name")
```
Dimosphen1, line 40: By general practice, all operations with model objects should be performed via a special `Manager` class that is accessible via the class attribute `objects`. In case you want to add the functionality that is related to a model object operations, you should create the new `Manager` class and extend it there. However, it is possible to use Mixin here because you add the same functionality to multiple classes, it is just a note so that you to remember what should be a first choice in such cases.

db/models.py
```diff
@@ -1 +1,31 @@
+from typing import List, Dict
+
 from django.db import models
+
+
+class BaseModel:
```
Dimosphen1, line 6: It would be better to name this class `BulkCreateMixin`.

Dimosphen1 (approved):
LGTM, several comments were left for consideration.

PR #510, opened 2023-09-19
Solution

Daniil-Pankieiev:


main.py
```diff
@@ -2,6 +2,48 @@
...
+        ("Keanu", "Reeves"),
+        ("Scarlett", "Keegan"),
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson")
+    ]
+    genres = ["Western", "Action", "Drama"]
+
+    for item_first, item_second in actors:
+        Actor.objects.create(
```
Familenko, line 20: You're very clever, you wrote down the first and last names separately)). But please label the variables unpacked from the tuple according to their contents

main.py
```diff
@@ -2,6 +2,48 @@
...
+            first_name=item_first,
+            last_name=item_second)
+
+    for item in genres:
+        Genre.objects.create(name=item)
```
Familenko, line 25: same issue here

main.py
```diff
@@ -2,6 +2,48 @@
...
+    ).update(last_name="Clooney")
+
+    Actor.objects.filter(
+        first_name="Kianu",
+        last_name="Reaves"
+    ).update(first_name="Keanu", last_name="Reeves")
+
+    Genre.objects.filter(
+        name="Action",
+    ).delete()
```
Familenko, line 41: please use same style in all code:

in this case you make a block

    Genre.objects.filter(
        name="Action",
    ).delete()

but in same situation you make one line
```
Genre.objects.filter(name="Dramma",).update(name="Drama")
```

Familenko (changes requested):

fsocie7y (approved):
GJ 👍

PR #521, opened 2023-09-30
Solution: Genres & Actors.

lobodInI:


db/models.py
```diff
@@ -1 +1,16 @@
 from django.db import models
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255)
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
viktoria-rybenchuk, line 8: Don't use an f-string if there is no need.

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #526, opened 2023-10-04
solution

Familenko:


main.py
```diff
@@ -2,6 +2,35 @@
...
+    ]
+
+    for genre in genres:
+        Genre.objects.create(name=genre)
+
+    actors = [
+        "George Klooney", "Kianu Reaves",
+        "Scarlett Keegan", "Will Smith",
+        "Jaden Smith", "Scarlett Johansson"
+    ]
```
prochigor, line 21: You don't need to use a list when you can have a tuple it is more optimized if you don't change data

fsocie7y (approved):
well done ^_^

PR #530, opened 2023-10-08
'Solution'

hanovskyiv:


viktoria-rybenchuk (changes requested):
you should add migrations to github as well

pavlejviki (approved):
Good job!

PR #537, opened 2023-10-18
Solution

asdadaversa:


asdadaversa:
its was problem with non-commit migration file

db/models.py
```diff
@@ -1 +1,10 @@
 from django.db import models
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255)
+
+
+class Actor(models.Model):
```

db/models.py
```diff
@@ -1 +1,16 @@
 from django.db import models
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255)
+
+    def __str__(self) -> None:
+        print(f"{self.name}")
```
Dimosphen1, line 8: You should place a `return` statement, not print in `__str__`. Indicate this in type annotation and do not overuse the f-strings, there is no need for them when you return `CharField` directly.

db/models.py
```diff
@@ -1 +1,16 @@
...
+
+
+class Actor(models.Model):
+    first_name = models.CharField(max_length=255)
+    last_name = models.CharField(max_length=255)
+
+    def __str__(self) -> str:
+        print(f"{self.first_name}, {self.last_name}")
```
Dimosphen1, line 16: You should use `return` instead of `print`.

Dimosphen1 (changes requested):
Several changes were requested.

db/models.py
asdadaversa: Which annotation type i should put here? Not None?

db/models.py
```diff
@@ -1 +1,16 @@
 from django.db import models
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255)
+
+    def __str__(self) -> None:
```
viktoria-rybenchuk, line 7: does this function return `None`? Fix  in all places

db/migrations/0001_initial.py
```diff
@@ -0,0 +1,29 @@
+# Generated by Django 4.0.2 on 2023-10-18 08:54
+
```
viktoria-rybenchuk, line 2: remove blank line

viktoria-rybenchuk (changes requested):

db/models.py
asdadaversa: i set up models.Charfield, if its also incorrect - help me pls)

pavlejviki (approved):
Well done!

db/models.py
```diff
@@ -1 +1,16 @@
...
+
+    def __str__(self) -> models.CharField:
+        return self.name
+
+
+class Actor(models.Model):
+    first_name = models.CharField(max_length=255)
+    last_name = models.CharField(max_length=255)
+
+    def __str__(self) -> models.CharField:
```
pavlejviki, line 15: sorry haven't noticed.

wrong type annotation. this method returns str

pavlejviki (changes requested):

db/models.py
asdadaversa: done, buy i already received comment about this:

You should place a return statement, not print in __str__. Indicate this in type annotation and do not overuse the f-strings, there is no need for them when you return CharField directly.

PR #560, opened 2023-10-30
solution

StepanShevchuk:


main.py
```diff
@@ -2,6 +2,29 @@
...
 from django.db.models import QuerySet
 
+from db.models import Actor, Genre
+
 
 def main() -> QuerySet:
-    pass
+    Genre.objects.create(name="Western")
+    Genre.objects.create(name="Action")
+    Genre.objects.create(name="Dramma")
```
ArsenPidhoretskyi, line 11: Use loop instead of duplicate code

main.py
```diff
@@ -2,6 +2,29 @@
...
+
+    Actor.objects.create(first_name="George", last_name="Klooney")
+    Actor.objects.create(first_name="Kianu", last_name="Reaves")
+    Actor.objects.create(first_name="Scarlett", last_name="Keegan")
+    Actor.objects.create(first_name="Will", last_name="Smith")
+    Actor.objects.create(first_name="Jaden", last_name="Smith")
+    Actor.objects.create(first_name="Scarlett", last_name="Johansson")
+    Genre.objects.filter(name="Dramma").update(name="Drama")
```
ArsenPidhoretskyi, line 19: Use loop instead of duplicate code

main.py
```diff
@@ -2,6 +2,29 @@
...
+    (Actor.objects.filter(first_name="George").filter(last_name="Klooney")
+     .update(last_name="Clooney"))
```
ArsenPidhoretskyi, line 21: brackets are redundant. Also no need to use two filters. Fix in all places
```suggestion
    Actor.objects.filter(
        first_name="George", last_name="Klooney"
    ).update(last_name="Clooney"))
```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #564, opened 2023-10-31
Solution

mykolamateichuk:


Shatrovskyi (approved):

LiudmylaKulinchenko (changes requested):
All checks should pass

main.py
```diff
@@ -1,7 +1,64 @@
...
+    for genre in genres:
+        Genre.objects.create(name=genre)
+
+    for actor in actors:
+        Actor.objects.create(
+            first_name=actor[0],
+            last_name=actor[1]
+        )
+
+    # Update
```
LiudmylaKulinchenko, line 34: You can remove comments

main.py
```diff
@@ -1,7 +1,64 @@
...
+    Genre.objects.filter(
+        name="Action"
+    ).delete()
+
+    Actor.objects.filter(
+        first_name="Scarlett"
+    ).delete()
+
+    # Return
+    return Actor.objects.filter(
```
LiudmylaKulinchenko, line 62: No need to split in lines if it fits in one

main.py
```diff
@@ -1,7 +1,64 @@
...
+        "Dramma"
+    ]
+
+    actors = [
+        ("George", "Klooney"),
+        ("Kianu", "Reaves"),
+        ("Scarlett", "Keegan"),
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson")
```
LiudmylaKulinchenko, line 21: better to use sets instead of lists, will make it more optimized

LiudmylaKulinchenko (changes requested):

PR #566, opened 2023-10-31
work with models

DenPrislipskyi:


main.py
```diff
@@ -1,7 +1,34 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Actor, Genre
 
 
 def main() -> QuerySet:
-    pass
+    genres = ["Western", "Action", "Dramma"]
```
prochigor, line 8: if your array is not changing, better to use a tuple, it is more optimized

prochigor (approved):

main.py
LiudmylaKulinchenko: I agree

LiudmylaKulinchenko (changes requested):

main.py
```diff
@@ -1,7 +1,34 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Actor, Genre
 
 
 def main() -> QuerySet:
-    pass
+    genres = ("Western", "Action", "Dramma")
+    actors = ("George Klooney", " Kianu Reaves", "Scarlett Keegan",
```
fsocie7y, line 9: why not create a list of tuples? you will not need to use split later

fsocie7y (changes requested):

fsocie7y (approved):
well done ^_^

PR #569, opened 2023-11-01
Creating and editing my first database

SelFisher1488:


OlehShumov (approved):
thanks for fixing it)

main.py
```diff
@@ -1,7 +1,35 @@
...
 
 def main() -> QuerySet:
-    pass
+    genres = ["Western", "Action", "Dramma"]
+    actors_names = [("George", "Klooney"),
+                    ("Kianu", "Reaves"),
+                    ("Scarlett", "Keegan"),
+                    ("Will", "Smith"),
+                    ("Jaden", "Smith"),
+                    ("Scarlett", "Johansson")]
```
LiudmylaKulinchenko, line 14: You can use tuples here instead, it will be more optimized

LiudmylaKulinchenko (changes requested):

PR #570, opened 2023-11-01
solution

kostomeister:


fedorov-s-od (approved):
Ти таким цікавим шляхом пішов, через створення та відкриття додаткового файлу )) Але працює! Гарна робота ))

.gitignore
```diff
@@ -7,4 +7,4 @@ venv/
 .pytest_cache/
 **__pycache__/
 *.pyc
-**db.sqlite3
+**db.sqlite3
```
LiudmylaKulinchenko, line 10: add a blank line

main.py
```diff
@@ -2,6 +2,31 @@
...
+        data = line.split(" ")
+
+        kind_of_data = data[0]
+
+        if kind_of_data == "genre":
+            genre = data[1]
+            Genre.objects.create(name=genre)
+        else:
+            first_name, last_name = data[1], data[2]
+            Actor.objects.create(first_name=first_name, last_name=last_name)
```
LiudmylaKulinchenko, line 22: It's interesting choice, but I imagine how you wroted those file manually 😣 You should avoid additional work. Will be much more optimized just to create tuples with data inside a function. Thats what you should think about: to make it more optimized and to do less work)

LiudmylaKulinchenko (changes requested):

main.py
kostomeister: it was actually harder to use tuples instead of a file :)

main.py
```diff
@@ -2,6 +2,34 @@
 
 from django.db.models import QuerySet
 
+from db.models import Actor, Genre
+
 
 def main() -> QuerySet:
-    pass
+    genres = ("Western", "Action", "Dramma")
+    actors = (
```
fsocie7y, line 10: why not create a list of tuples?) you will not need to use split later)

fsocie7y (approved):

PR #572, opened 2023-11-01
Solution Cinema Hall

Oxbay:


Lyutillis (changes requested):
Try running these commands:
    py manage.py makemigrations
    py manage.py migrate
To make db migrations.
And then add the whole folder`s changes to commit by using
    git add .

fsocie7y (approved):
well done ^_^

PR #573, opened 2023-11-01
Solution

OlehShumov:


RabenkoYevhenii (approved):

main.py
```diff
@@ -1,7 +1,37 @@
...
+        "Dramma"
+    ]
+    actors = [
+        ("George", "Klooney"),
+        ("Kianu", "Reaves"),
+        ("Scarlett", "Keegan"),
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson")
+    ]
```
LiudmylaKulinchenko, line 20: You can use tuples here instead, it will be more optimized

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌷

PR #575, opened 2023-11-01
Solution

skoval7:


kostomeister (changes requested):
add migration file to get rid of an error

LiudmylaKulinchenko (approved):
Good job 🌷

PR #578, opened 2023-11-01
Solution

katyakoziar:


SevKrok (approved):
Nice work ;)

main.py
```diff
@@ -1,7 +1,47 @@
...
-    pass
+    genres = ["Western", "Action", "Dramma"]
+    actors = [
+        ("George", "Klooney"),
+        ("Kianu", "Reaves"),
+        ("Scarlett", "Keegan"),
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson")
+    ]
```
LiudmylaKulinchenko, line 16: You can use tuples here instead, it will be more optimized

main.py
```diff
@@ -1,7 +1,47 @@
...
+
+    for genre in genres:
+        Genre.objects.create(
+            name=genre,
+        )
+
+    for (first_name, second_name) in actors:
```
LiudmylaKulinchenko, line 23: You can skip parentheses here

LiudmylaKulinchenko (changes requested):

PR #579, opened 2023-11-01
Solution

MiskoRuslan:


Katherine-Greg (approved):

main.py
```diff
@@ -1,7 +1,35 @@
...
 
 from django.db.models import QuerySet
+from db.models import Actor, Genre
 
 
 def main() -> QuerySet:
-    pass
+    genres = ("Western", "Action", "Dramma")
+
+    actors = ("George Klooney", " Kianu Reaves", "Scarlett Keegan",
```
fsocie7y, line 10: why not create a list of tuples? you will not need to use split later)

fsocie7y (changes requested):

main.py
MiskoRuslan: done 🟢

LiudmylaKulinchenko (approved):
Keep it up! 🔥

PR #582, opened 2023-11-01
Develop

Andrii-Silenko:


OlehShumov (approved):
maybe like this - `return Actor.objects.filter(last_name="Smith").order_by("first_name")` is easier to read.

main.py
```diff
@@ -1,7 +1,63 @@
...
-    pass
+    genres = ["Western", "Action", "Dramma"]
+    actors = [
+        {"first_name": "George", "last_name": "Klooney"},
+        {"first_name": "Kianu", "last_name": "Reaves"},
+        {"first_name": "Scarlett", "last_name": "Keegan"},
+        {"first_name": "Will", "last_name": "Smith"},
+        {"first_name": "Jaden", "last_name": "Smith"},
+        {"first_name": "Scarlett", "last_name": "Johansson"},
+    ]
```
LiudmylaKulinchenko, line 16: use set instead to make more optimized

main.py
```diff
@@ -1,7 +1,63 @@
...
+        name="Action"
+    ).delete()
+
+    Actor.objects.filter(
+        first_name="Scarlett"
+    ).delete()
+
+    return Actor.objects.filter(
+        last_name="Smith"
+    ).order_by("first_name")
```
LiudmylaKulinchenko, line 59: no need to split in lines if it fits in one

LiudmylaKulinchenko (changes requested):

PR #592, opened 2023-11-10
Solution

Sparix:


main.py
```diff
@@ -1,7 +1,48 @@
...
-    pass
+    genres = ["Western", "Action", "Dramma"]
+    actors = [
+        "George Klooney",
+        "Kianu Reaves",
+        "Scarlett Keegan",
+        "Will Smith",
+        "Jaden Smith",
+        "Scarlett Johansson"
+    ]
```
viktoria-rybenchuk, line 16: save name and last name separately and then you can use unpacking. For example `("George",  "Klooney")`. Your code will be more efficient and readable

main.py
```diff
@@ -1,7 +1,48 @@
...
+    Actor.objects.filter(
+        first_name="Kianu",
+        last_name="Reaves"
+    ).update(first_name="Keanu", last_name="Reeves")
+
+    Genre.objects.filter(name="Action").delete()
+
+    Actor.objects.filter(first_name="Scarlett").delete()
+
+    all_actor = Actor.objects.filter(
```
viktoria-rybenchuk, line 45: why do yon need variable name?

viktoria-rybenchuk (changes requested):

db/models.py
```diff
@@ -1 +1,16 @@
 from django.db import models
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255)
+
+    def __str__(self) -> str:
+        return f"{self.name}"
```
pavlejviki, line 8: no need to use f-string

pavlejviki (approved):
Good job!

PR #605, opened 2023-11-21
Solution

sachavskyi:


Polyakiv-Andrey (changes requested):
Test not passed!

db/models.py
```diff
@@ -1 +1,10 @@
 from django.db import models
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255)
+
+
+class Actor(models.Model):
+    first_name = models.CharField(max_length=255)
+    last_name = models.CharField(max_length=255)
```
pavlejviki, line 10: add str method to models

main.py
```diff
@@ -1,7 +1,35 @@
...
-    pass
+    Genre.objects.create(name="Western")
+    Genre.objects.create(name="Action")
+    Genre.objects.create(name="Dramma")
+    Actor.objects.create(first_name="George", last_name="Klooney")
+    Actor.objects.create(first_name="Kianu", last_name="Reaves")
+    Actor.objects.create(first_name="Scarlett", last_name="Keegan")
+    Actor.objects.create(first_name="Will", last_name="Smith")
+    Actor.objects.create(first_name="Jaden", last_name="Smith")
+    Actor.objects.create(first_name="Scarlett", last_name="Johansson")
```
pavlejviki, line 16: you should optimise your code to avoid duplication

use a list of tuples that include the first name and last name of each actor and iterate over that list passing the correct values to the create method of the Actor model

pavlejviki (changes requested):

pavlejviki (approved):
Good job!

PR #606, opened 2023-11-22
solution_py-genres-and-actors_1

VladKrit-commits:


db/models.py
```diff
@@ -1 +1,10 @@
 from django.db import models
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255)
+
+
+class Actor(models.Model):
+    first_name = models.CharField(max_length=255, null=False)
+    last_name = models.CharField(max_length=255, null=False)
```
pavlejviki, line 10: add str methods to your models

main.py
```diff
@@ -1,7 +1,36 @@
...
-    pass
+    genre_list = ["Western", "Action", "Dramma"]
+    actors_list = [
+        ["George", "Klooney"],
+        ["Kianu", "Reaves"],
+        ["Scarlett", "Keegan"],
+        ["Will", "Smith"],
+        ["Jaden", "Smith"],
+        ["Scarlett", "Johansson"]
+    ]
```
pavlejviki, line 16: Using a list of tuples is better.

Tuples are often used to represent fixed collections of items. In your case, each actor's name is a fixed pair of first and last names. Using tuples can make it clearer that these pairs are fixed sets of data.

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #621, opened 2023-12-05
solution

oleksashcherbakov:


main.py
```diff
@@ -1,7 +1,36 @@
...
 
 def main() -> QuerySet:
-    pass
+    genres = ["Western", "Action", "Dramma"]
+    actors = ["George Klooney",
+              "Kianu Reaves",
+              "Scarlett Keegan",
+              "Will Smith",
+              "Jaden Smith",
+              "Scarlett Johansson"]
```
Dimosphen1, line 14: You could place tuples with first name and last name inside of the list to use unpacking during the iteration. It will reduce the need to make a `split` operation on line 20. Refactor your code to achieve that.

main.py
```diff
@@ -1,7 +1,36 @@
...
+
+    for genre in genres:
+        Genre.objects.create(name=genre)
+
+    for actor in actors:
+        first_name, last_name = actor.split()
+        Actor.objects.create(
+            first_name=first_name,
+            last_name=last_name)
```
Dimosphen1, line 23: In case you do not have any arguments on the same line as the closing parenthesis, your opening parenthesis should not have any arguments on the same line as well. Reformat your code to encounter that.

main.py
```diff
@@ -1,7 +1,36 @@
...
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+
+    Actor.objects.filter(last_name="Klooney").update(last_name="Clooney")
+    (Actor.objects.filter(first_name="Kianu", last_name="Reaves").
+     update(first_name="Keanu", last_name="Reeves"))
```
Dimosphen1, line 29: There is no need to place these expressions inside of `()`, make the line break after the opening parenthesis for `filter` or in the other place. Make this change to all places where you put the expression inside of `()`.

main.py
```diff
@@ -1,7 +1,36 @@
...
+
+    Genre.objects.filter(name="Action").delete()
+    Actor.objects.filter(first_name="Scarlett").delete()
+
+    ordered_actors = (Actor.objects.filter(last_name="Smith").
+                      order_by("first_name"))
+    return ordered_actors
```
Dimosphen1, line 36: There is no need to allocate the `ordered_actors` variable, return the queryset directly.

Dimosphen1 (changes requested):
Several changes were requested.

oleksashcherbakov:
fixed

main.py
```diff
@@ -1,7 +1,38 @@
...
+        ("Scarlett", "Keegan"),
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson")
+    ]
+
+    for genre in genres:
+        Genre.objects.create(name=genre)
+
+    for actor in actors:
```
viktoria-rybenchuk, line 21: Unpack  the actor tuple into `first_name` and `last_name` directly in the loop. This makes the code more readable and eliminates the need for indexing when accessing the elements of the actor tuple.

viktoria-rybenchuk (changes requested):

main.py
oleksashcherbakov: one more time

pavlejviki (approved):
Well done!

PR #629, opened 2023-12-13
Solution

Lebwa1337:


Lebwa1337:
Also, i have a question.What are these warnings?
image

PR #642, opened 2024-01-10
'Solution_Genres-and-actors'

vvg04021968:


main.py
```diff
@@ -2,6 +2,37 @@
...
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
-    pass
+    genres = [
+        ("Western",),
+        ("Action",),
+        ("Dramma",)
+    ]
```
pavlejviki, line 13: There's no need for genres to be a list of tuples if you're only using a single field (name) for the Genre model. it can be a simple list of names

pavlejviki (changes requested):

vvg04021968:
I will sign up for Q&A  / I have a question
--All checks have failed--

main.py
```diff
@@ -2,6 +2,33 @@
 
 from django.db.models import QuerySet
 
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
-    pass
+    genres = ["Western", "Action", "Dramma"]
+    for genre, in genres:
```
pavlejviki, line 10: remove the comma

pavlejviki (changes requested):

vasylhnatiuk (approved):

PR #671, opened 2024-02-07
Develop

zdmytryszyn:


db/models.py
```diff
@@ -1 +1,16 @@
 from django.db import models
+
+
+class Genre(models.Model):
+    name = models.CharField(max_length=255)
+
+    def __repr__(self) -> str:
```
Polyakiv-Andrey, line 7: it would be better if you implemented the str method instead of repr

db/models.py
```diff
@@ -1 +1,16 @@
...
+        return f"<Genre: {self.name}>"
+
+
+class Actor(models.Model):
+    first_name = models.CharField(max_length=255)
+    last_name = models.CharField(max_length=255)
+
+    def __repr__(self) -> str:
```
Polyakiv-Andrey, line 15: it would be better if you implemented the str method instead of repr

Polyakiv-Andrey (changes requested):

vasylhnatiuk (approved):

PR #690, opened 2024-02-09
Solution

oksankakot:


Serejka21 (approved):
Great Job! I been talking with Oksana and she should change her main function in create() method case, using list of values and loop

PR #715, opened 2024-02-22
Solution

vsupruniuk:
image

One of the test fails, but the only difference is the order of the actors


main.py
```diff
@@ -2,6 +2,43 @@
...
+    ).update(last_name="Clooney")
+    Actor.objects.filter(
+        first_name="Kianu",
+        last_name="Reaves"
+    ).update(first_name="Keanu", last_name="Reeves")
+
+    Genre.objects.filter(name="Action").delete()
+    Actor.objects.filter(first_name="Scarlett").delete()
+
+    return Actor.objects.filter(last_name="Smith").order_by("first_name").all()
```
LiudmylaKulinchenko, line 40: this code will work the same without `.all()`

LiudmylaKulinchenko (changes requested):
There is another error on github, not about actors order

LiudmylaKulinchenko (approved):
Good job 🔥

PR #732, opened 2024-03-05
Solution

AlexTsikhun:
I think,  there is not problem that migrations is 2?
 Actors list with tuples is redable? or better to use 1 tuple in row (because code writes from top to down... and so on)?


db/migrations/0002_actor_remove_genre_first_name_remove_genre_last_name.py
```diff
@@ -0,0 +1,29 @@
...
+        ),
+        migrations.RemoveField(
+            model_name='genre',
+            name='first_name',
+        ),
+        migrations.RemoveField(
+            model_name='genre',
+            name='last_name',
+        ),
+    ]
```
Dimosphen1, line 29: It is a good practice to include one migration per app per PR in Django projects since it helps to avoid `migrations` folder overload in the future. Represent all changes done to models in one migration file per app per PR, please

Dimosphen1 (changes requested):

vsmutok (approved):
Looks Good!

PR #767, opened 2024-04-03
done?

antonver:


main.py
```diff
@@ -2,6 +2,28 @@
...
+         Genre(name="Action"),
+         Genre(name="Dramma"),
+         ]
+    )
+    Actor.objects.bulk_create(
+        [Actor(first_name="George", last_name="Klooney"),
+         Actor(first_name="Kianu", last_name="Reaves"),
+         Actor(first_name="Scarlett", last_name="Keegan"),
+         Actor(first_name="Will", last_name="Smith"),
+         Actor(first_name="Jaden", last_name="Smith"),
```
Dimosphen1, line 20: It would be better to define two lists with values and populate actors and genres in a loop from there to avoid code duplication. You could also make a list comprehension while passing new instances to the `bulk_create`.

EdAlekseiev (approved):

PR #780, opened 2024-04-11
solution

Eugen-Veklich:


main.py
```diff
@@ -2,6 +2,29 @@
 
 from django.db.models import QuerySet
 
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
-    pass
+    Genre.objects.create(name="Western")
```
vasylhnatiuk, line 9: use loop instead of duplicating

main.py
```diff
@@ -2,6 +2,29 @@
...
+    Genre.objects.create(name="Action")
+    Genre.objects.create(name="Dramma")
+    Actor.objects.create(first_name="George", last_name="Klooney")
```
vasylhnatiuk, line 12: use loop instead of duplicating

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #790, opened 2024-04-17
solution

ArtemLeo:


main.py
```diff
@@ -1,7 +1,47 @@
...
+    Actor.objects.filter(
+        first_name="Kianu",
+        last_name="Reaves"
+    ).update(
+        first_name="Keanu",
+        last_name="Reeves"
+    )
+
+    Genre.objects.filter(name="Action").delete()
+    Actor.objects.filter(first_name__startswith="Scarlett").delete()
```
Dimosphen1, line 45: There is no need to explicitly indicate `__startswith` since task asks for an exact match. Remove this part from the ORM

main.py
```diff
@@ -1,7 +1,47 @@
...
 
 
+from db.models import Genre, Actor
+
+
 def main() -> QuerySet:
     pass
+    genres_data = [("Western",), ("Action",), ("Drama",)]
+
+    for genre_name, in genres_data:
```
Dimosphen1, line 13: There is no need to create a list of tuples since they contain only one string. Create a list of strings instead and iterate through them

main.py
```diff
@@ -1,7 +1,47 @@
...
 
 from django.db.models import QuerySet
 
 
+from db.models import Genre, Actor
+
+
 def main() -> QuerySet:
     pass
+    genres_data = [("Western",), ("Action",), ("Drama",)]
```
Dimosphen1, line 11: You should create genre `Dramma`, not `Drama` here

main.py
```diff
@@ -1,7 +1,47 @@
...
+        ("Scarlett", "Keegan"),
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson")
+    ]
+
+    for first_name, last_name in actors_data:
+        Actor.objects.create(first_name=first_name, last_name=last_name)
+
+    Genre.objects.filter(name="Drama").update(name="Drama")
```
Dimosphen1, line 28: This code should be the following
```suggestion
    Genre.objects.filter(name="Dramma").update(name="Drama")
```

main.py
```diff
@@ -1,7 +1,47 @@
...
+
+    Actor.objects.filter(
+        first_name="George",
+        last_name="Klooney"
+    ).update(
+        last_name="Clooney")
```
Dimosphen1, line 34: In case there are no arguments on the same line with an opening parenthesis, your closing parenthesis should not have any arguments on the same line as well. Please, resolve this issue

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #791, opened 2024-04-17
Solution

Gilgumish:


main.py
```diff
@@ -1,7 +1,36 @@
...
+    actors = [
+        ("George", "Klooney"),
+        ("Kianu", "Reaves"),
+        ("Scarlett", "Keegan"),
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson")
+    ]
+
+    for genre in genres:
```
Oleksl888, line 19: There is no mistake. But here would make sense to break down the function into smaller function to follow SRP.

Oleksl888 (approved):
Good job!

PR #797, opened 2024-04-17
Solution

SashaKisliy:


main.py
```diff
@@ -1,7 +1,41 @@
...
+            Actor.objects.create(first_name=item[1],
+                                 last_name=item[2])
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+    Actor.objects.filter(last_name="Klooney").update(last_name="Clooney")
+
+    (Actor.objects.filter(
+        first_name="Kianu", last_name="Reaves")
+     .update(
+        first_name="Keanu", last_name="Reeves"))
```
YuriiKindrat, line 30: bad formatting here

main.py
```diff
@@ -1,7 +1,41 @@
...
+            ("actress", "Scarlett", "Keegan"),
+            ("actor", "Will", "Smith"),
+            ("actor", "Jaden", "Smith"),
+            ("actress", "Scarlett", "Johansson")]
+    for item in data:
+        if item[0] == "genre":
+            Genre.objects.create(name=item[1])
+        if item[0] == "actor" or item[0] == "actress":
+            Actor.objects.create(first_name=item[1],
+                                 last_name=item[2])
```
YuriiKindrat, line 22: why dont u separate the data e.g. genres/actors and make a 2 separate loop for creating for better readability ?

YuriiKindrat (changes requested):

main.py
SashaKisliy: previously, I was told that it's preferable not to create multiple loops if everything can be done in one

Oleksl888: here is ok to have two loops since these are different entities and combining them into one loop will not affect performance. here you have 3 iterations for genres and 6 iterations for actors. Wether you put them in one loop or separate it will still be 9 iterations total

main.py
```diff
@@ -1,7 +1,36 @@
...
+              ("Kianu", "Reaves"),
+              ("Scarlett", "Keegan"),
+              ("Will", "Smith"),
+              ("Jaden", "Smith"),
+              ("Scarlett", "Johansson")]
+    for first_name, last_name in actors:
+        Actor.objects.create(first_name=first_name, last_name=last_name)
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+    Actor.objects.filter(last_name="Klooney").update(last_name="Clooney")
```
Oleksl888, line 21: here would be more appropriate to filter by first and last name to get more precise results

main.py
```diff
@@ -1,7 +1,36 @@
...
+
+    (Actor.objects.
```
Oleksl888, line 23: here would be better to follow this:
```
Actor.objects.filter(
   ...
).update(
   ...

Oleksl888 (changes requested):

PR #801, opened 2024-04-18
solution

alonasorochynska:


main.py
```diff
@@ -2,6 +2,68 @@
...
+        ["Jaden", "Smith"],
+        ["Scarlett", "Johansson"]
+    )
+    for genre in genres:
+        Genre.objects.create(
+            name=genre
+        )
+
+    for actor in actors:
+        Actor.objects.create(
```
fsocie7y, line 29: You can unpack lists in place, if you want. For example,
for value1, value2 in list_of_pairs:

main.py
```diff
@@ -2,6 +2,68 @@
...
+    ]
+
+    actors = (
+        ["George", "Klooney"],
+        ["Kianu", "Reaves"],
+        ["Scarlett", "Keegan"],
+        ["Will", "Smith"],
+        ["Jaden", "Smith"],
+        ["Scarlett", "Johansson"]
+    )
```
fsocie7y, line 22: why not create a list of tuples? you will not need to use split later

fsocie7y (changes requested):

main.py
```diff
@@ -2,6 +2,68 @@
...
+            name=genre
+        )
+
+    for first_name, last_name in actors:
+        Actor.objects.create(
+            first_name=first_name,
+            last_name=last_name,
+        )
+
+    Genre.objects.all().filter(
```
YuriiKindrat, line 34: no need to use all before filter )

YuriiKindrat (approved):
GJ

PR #821, opened 2024-04-25
solution

haldaniko:


main.py
```diff
@@ -1,7 +1,38 @@
...
+    )
+    for genre in genres:
+        Genre.objects.create(
+            name=genre
+        )
+    for actor in actors:
+        Actor.objects.create(
+            first_name=actor[0],
+            last_name=actor[1]
+        )
```
EdAlekseiev, line 25: Here you should use unboxing, it will make your code more readable
```suggestion
    for first_name, last_name  in actors:
        Actor.objects.create(
            first_name=first_name,
            last_name=last_name
        )
```

EdAlekseiev (changes requested):

main.py
haldaniko: Thanks!

EdAlekseiev (changes requested):
Fix flake8, please

EdAlekseiev (approved):

PR #824, opened 2024-04-27
Solution

AndriiDraga:


db.sqlite3
Dimosphen1, line 1: Remove a DB (database) from the PR (pull request), it should not be included

main.py
```diff
@@ -2,6 +2,40 @@
...
 
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
-    pass
+    genres = [
+        ("Western",),
+        ("Action",),
+        ("Dramma",),
```
Dimosphen1, line 12: There is no need to add genres in tuples and after that index them during retrieval, place them as strings inside of a list

Dimosphen1 (changes requested):

EdAlekseiev (approved):

PR #866, opened 2024-06-14
solution

AndrewwSSS:


main.py
```diff
@@ -1,7 +1,65 @@
...
-    pass
+    Genre.objects.create(
+        name="Western"
+    )
+    Genre.objects.create(
+        name="Action"
+    )
+    Genre.objects.create(
+        name="Dramma"
+    )
```
DevSpaciX, line 16: DRY ! Look you can create list of genres and then create them in db using for loop.

main.py
```diff
@@ -1,7 +1,65 @@
...
+        last_name="Smith"
+    )
+    Actor.objects.create(
+        first_name="Jaden",
+        last_name="Smith"
+    )
+    Actor.objects.create(
+        first_name="Scarlett",
+        last_name="Johansson"
+    )
```
DevSpaciX, line 40: Same here . Do not repeat yourself

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #871, opened 2024-06-18
Solution_v1

BileichukIvan:


main.py
```diff
@@ -1,7 +1,37 @@
...
+from db.models import Genre, Actor
 
 
 def main() -> QuerySet:
-    pass
+    genres = [
+        Genre(name="Western"),
+        Genre(name="Action"),
+        Genre(name="Dramma"),
+    ]
```
DevSpaciX, line 12: That is not good practice to create dataset with instances of some model . It is better to create just a list or tuples if you need to store more data in . And then create db data using `create() `
Bulk create also good too , but on really big sets of data ( 100 + rows for example )

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #874, opened 2024-06-18
Solution

lilarin:


main.py
```diff
@@ -1,7 +1,32 @@
...
+
+    actors = [
+        Actor(first_name="George", last_name="Klooney"),
+        Actor(first_name="Keanu", last_name="Reaves"),
+        Actor(first_name="Scarlett", last_name="Keegan"),
+        Actor(first_name="Will", last_name="Smith"),
+        Actor(first_name="Jaden", last_name="Smith"),
+        Actor(first_name="Scarlett", last_name="Johansson"),
+    ]
+    Actor.objects.bulk_create(actors)
```
DevSpaciX, line 22: That is not good practice to create dataset with instances of some model . It is better to create just a list or tuples if you need to store more data in . And then create db data using `create() `
Bulk create also good too , but on really big sets of data ( 100 + rows for example )

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #876, opened 2024-06-19
Solution

necobra:


main.py
```diff
@@ -2,6 +2,58 @@
...
+            name=genre_name
+        )
+
+    actors_names_to_create = [
+        "George Klooney", "Kianu Reaves", "Scarlett Keegan",
+        "Will Smith", "Jaden Smith", "Scarlett Johansson"
+    ]
+    for actor_name in actors_names_to_create:
+        actor_name_dict = {"first_name": actor_name.split()[0],
+                           "last_name": actor_name.split()[1]}
```
DevSpaciX, line 22: why not create a list of tuples? you will not need to use split later

main.py
```diff
@@ -2,6 +2,58 @@
...
+            "first_name":
+                actors_to_update["names_to_update"][index].split()[0],
+            "last_name":
+                actors_to_update["names_to_update"][index].split()[1]
+        }
+        Actor.objects.filter(
+            **actor_name_dict
+        ).update(
+            **actors_to_update["new_atrbs"][index]
+        )
```
DevSpaciX, line 53: Just do it by single names . We do not need dict fot updating data here

DevSpaciX (changes requested):

main.py
necobra: i thought about the idea that we input the full name and next create an actor. So, the creation of a list of tuples just adds one step more

main.py
necobra: i just wanted to implement more auto process for updating, but I agree that was horrible)

main.py
DevSpaciX: Why it will add one more step ? Your data will be already separated be first and last name in each tuple . For example --- >
```
actors_data = [
    ( first_name, last_name),
    (... , ... )
]
```
Then --- >

`for first_name , last_name in actors_data:`

main.py
```diff
@@ -2,6 +2,47 @@
...
+
+    input_actors_fullnames_to_create = [
+        "George Klooney", "Kianu Reaves", "Scarlett Keegan",
+        "Will Smith", "Jaden Smith", "Scarlett Johansson"
+    ]
+    actors_names_to_create = (
+        (actor_name.split()[0], actor_name.split()[1])
+        for actor_name
+        in input_actors_fullnames_to_create
+    )
```
DevSpaciX, line 24: I left the comment about it . You can try it 
For example --- >
```
actors_data = [
    ( first_name, last_name),
    (... , ... )
]
```
Then --- >

`for first_name , last_name in actors_data:`

DevSpaciX (changes requested):

DevSpaciX (approved):
Good , next time delete all comments before push

PR #883, opened 2024-06-19
new Solution

SrMusienko:


Abnormaltype (changes requested):
Reject

Abnormaltype (approved):
use `black`  for migrations as well

PR #893, opened 2024-06-20
sol1

IvanVahanov:


main.py
```diff
@@ -1,7 +1,45 @@
...
+    Genre.objects.create(name="Western")
+    Genre.objects.create(name="Action")
+    Genre.objects.create(name="Dramma")
+
+    Actor.objects.create(first_name="George", last_name="Klooney")
+    Actor.objects.create(first_name="Kianu", last_name="Reaves")
+    Actor.objects.create(first_name="Scarlett", last_name="Keegan")
+    Actor.objects.create(first_name="Will", last_name="Smith")
+    Actor.objects.create(first_name="Jaden", last_name="Smith")
+    Actor.objects.create(first_name="Scarlett", last_name="Johansson")
```
DevSpaciX, line 17: Good practive will be :
1. Create dataset with  data  of future genres / actors
2. In for loop call `create()` method with needed name or last_name

DevSpaciX (changes requested):

PR #898, opened 2024-06-21
solution

DimaYankovskiy:


main.py
```diff
@@ -2,6 +2,28 @@
...
 
 def main() -> QuerySet:
-    pass
+    genres = ["Western", "Action", "Drama"]
+    for genre in genres:
+        Genre.objects.create(name=genre)
+
+    actors = ["George Clooney", "Keanu Reeves", "Will Smith", "Jaden Smith"]
+    for actor in actors:
+        first_name, last_name = actor.split()
```
DevSpaciX, line 15: why not create a list of tuples? you will not need to use split later

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #899, opened 2024-06-21
Solution

YevaYe:


main.py
```diff
@@ -2,6 +2,31 @@
...
 def main() -> QuerySet:
-    pass
+    genres = ["Western", "Action", "Dramma"]
+    for genre in genres:
+        Genre.objects.create(name=genre)
+
+    first_name = ["George", "Kianu", "Scarlett", "Will", "Jaden", "Scarlett"]
+    last_name = ["Klooney", "Reaves", "Keegan", "Smith", "Smith", "Johansson"]
+
+    for first, last in zip(first_name, last_name):
```
DevSpaciX, line 16: why not create a list of tuples? you will not need to use zip() later
Also naming `first` and `last` is not very informative

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #901, opened 2024-06-22
Solution

VitaliiDats1:


main.py
```diff
@@ -1,7 +1,31 @@
...
 def main() -> QuerySet:
-    pass
+    genres = ["Western", "Action", "Dramma"]
+    for genre_name in genres:
+        Genre.objects.create(name=genre_name)
+
+    first_name = ["George", "Kianu", "Scarlett", "Will", "Jaden", "Scarlett"]
+    last_name = ["Klooney", "Reaves", "Keegan", "Smith", "Smith", "Johansson"]
+    if len(first_name) != len(last_name):
+        raise ValueError("Value error")
```
DevSpaciX, line 15: Why did you add this check ? You can create just one list with tuples inside . For example [(first_name , last_name ), ...] . It will be more cleaner

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #909, opened 2024-07-02
implement py-geners-and-actors

HalynaRohatska:


main.py
```diff
@@ -2,6 +2,48 @@
...
+    actors = [
+        ["George", "Klooney"],
+        ["Kianu", "Reaves"],
+        ["Scarlett", "Keegan"],
+        ["Will", "Smith"],
+        ["Jaden", "Smith"],
+        ["Scarlett", "Johansson"]
+    ]
+
+    for actor in actors:
```
Oleksl888, line 23: better use unpacking here instead of indexing

main.py
```diff
@@ -2,6 +2,48 @@
...
+    ).update(first_name="Keanu", last_name="Reeves")
+
+    Genre.objects.filter(name="Action").delete()
+
+    Actor.objects.filter(first_name="Scarlett").delete()
+
+    return Actor.objects.filter(last_name="Smith").order_by("first_name")
+
+
+if __name__ == "__main__":
```
Oleksl888, line 48: this is not as per checklist

main.py
```diff
@@ -2,6 +2,48 @@
...
+        )
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+
+    Actor.objects.filter(
+        first_name="George",
+        last_name="Klooney"
+    ).update(last_name="Clooney")
+
+    Actor.objects.filter(
```
Oleksl888, line 36: You can break down main function in smaller specialized functions to structure your code better. E.g separate function for creating, updating, deleting..

Oleksl888 (changes requested):

main.py
Oleksl888: Whatever is on checklist is non-negotiable
> 1. Don't add this fragment:
> if __name__ == "__main__":
>     main()

Oleksl888 (changes requested):

Oleksl888 (approved):
Good job!

PR #932, opened 2024-07-21
implemented "cinema hall" task

MasakDirt:


main.py
```diff
@@ -1,7 +1,64 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Genre, Actor
```

main.py
```diff
@@ -1,7 +1,64 @@
...
+
+    return Actor.objects.filter(last_name="Smith").order_by("first_name")
+
+
+def create_genres(genres_names: tuple) -> None:
+    for name in genres_names:
+        Genre.objects.create(name=name)
+
+
+def create_actors(actors: dict) -> None:
```
Oleksl888, line 61: you could pass actors as a list of tuples or tuple of tuples and use unpacking - would be more elegant like that

main.py
```diff
@@ -1,7 +1,64 @@
...
+        },
+
+        "actress2": {
+            "first_name": "Scarlett",
+            "last_name": "Johansson"
+        },
+    })
+
+    # update
+    Genre.objects.filter(name="Dramma").update(name="Drama")
```
Oleksl888, line 42: you could also define update/delete/retrieve as separate fuctions to better structure your code.
On the same note you can have one function for creating both genres and actors

Oleksl888 (changes requested):

main.py
```diff
@@ -1,7 +1,44 @@
...
+    (Actor.objects.filter(first_name="Kianu", last_name="Reaves")
+     .update(first_name="Keanu", last_name="Reeves"))
+
+
+def delete_genre_and_actor() -> None:
+    Genre.objects.filter(name="Action").delete()
+    Actor.objects.filter(first_name="Scarlett").delete()
+
+
+def actor_order() -> QuerySet:
```
Oleksl888, line 42: `get_actor_by_name` would be a better name

Oleksl888 (approved):
Well done!

PR #943, opened 2024-07-31
Solution

Lebid-Dmytro:


main.py
```diff
@@ -1,7 +1,42 @@
...
-
 from django.db.models import QuerySet
 
+import init_django_orm  # noqa: F401
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
-    pass
+    # Create genres
```
Dimosphen1, line 8: There is no need to add these comments, remove them

main.py
```diff
@@ -1,7 +1,42 @@
...
+    Genre.objects.create(name="Western")
+    Genre.objects.create(name="Action")
+    Genre.objects.create(name="Dramma")
```
Dimosphen1, line 11: Try to avoid code duplication while creating genres and actors. Prefer creating these objects in a loop while iterating over data structure with required values

main.py
```diff
@@ -1,7 +1,42 @@
...
+    Genre.objects.filter(name="Action").delete()
+
+    # Delete actresses named Scarlett
+    Actor.objects.filter(first_name="Scarlett").delete()
+
+    # Return QuerySet of actors with last_name "Smith" ordered by first_name
+    smith_actors = Actor.objects.filter(
+        last_name="Smith"
+    ).order_by("first_name")
+    return smith_actors
```
Dimosphen1, line 42: There is no need to create `smith_actors` variable, return QuerySet directly

Dimosphen1 (changes requested):

PR #947, opened 2024-08-01
Implemented solution

dimak20:


main.py
```diff
@@ -1,7 +1,53 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Genre, Actor
```

main.py
```diff
@@ -1,7 +1,53 @@
...
+
+
+genre_items = (
```
Oleksl888, line 7: Why not call it genre_names and actor_info or something more appropriate then items?
Also these are used as global variables in your code which is not a good practice - you either declare them inside main or pass to main() as arguments

main.py
```diff
@@ -1,7 +1,53 @@
...
+)
 
 
 def main() -> QuerySet:
-    pass
+    for genre in genre_items:
+        Genre.objects.create(
+            name=genre
+        )
+    for item_first, item_second in actor_items:
```
Oleksl888, line 27: why not call it  first_name / last_name?

main.py
```diff
@@ -1,7 +1,53 @@
...
+        last_name="Klooney"
+    ).update(last_name="Clooney")
+    Actor.objects.filter(
+        first_name="Kianu",
+        last_name="Reaves"
+    ).update(
+        first_name="Keanu",
+        last_name="Reeves"
+    )
+    Genre.objects.filter(
```
Oleksl888, line 45: Would make sense to break down main function is smaller specialized functions to structure your code better. E.g separate function for creating, updating, deleting..

Oleksl888 (changes requested):

Oleksl888 (approved):
You know I would be happy if you did more simple implementation of methods something like below. What you did is great - you basically parametrized everything but your functions have too many conditional branches and so it looks a bit ike spaghetti :)
```
def update_actor():
 Actor.objects.filter(
        first_name="George",
        last_name="Klooney",
    ).update(last_name="Clooney")
    Actor.objects.filter(
        first_name="Kianu",
        last_name="Reaves",
    ).update(first_name="Keanu", last_name="Reeves")

PR #954, opened 2024-08-10
"Solution"

OleksiukStepan:


main.py
```diff
@@ -1,7 +1,51 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Genre, Actor
```

main.py
```diff
@@ -1,7 +1,51 @@
...
+def create_actors() -> None:
+    actors = [
+        "George Klooney",
+        "Kianu Reaves",
+        "Scarlett Keegan",
+        "Will Smith",
+        "Jaden Smith",
+        "Scarlett Johansson"
+    ]
+    for actor in actors:
```
Oleksl888, line 29: you can use unpacking here - would be more pythonic and saves you one line of code :)

Oleksl888, line 29: you can use unpacking here - would be more pythonic and saves you one line of code :)

main.py

main.py
```diff
@@ -1,7 +1,51 @@
...
+        first_name="Kianu", last_name="Reaves"
+    ).update(first_name="Keanu", last_name="Reeves")
+
+
+def delete() -> None:
+    Genre.objects.filter(name="Action").delete()
+    Actor.objects.filter(first_name="Scarlett").delete()
+
+
+def specific_actors() -> QuerySet:
```
Oleksl888, line 49: `get_specific_actors` would be a better name

Oleksl888 (changes requested):
Good job! But please check last

main.py
```diff
@@ -2,6 +2,50 @@
...
+
+
+def create_genres() -> None:
+    for name in ["Western", "Action", "Dramma"]:
+        Genre.objects.create(name=name)
+
+
+def create_actors() -> None:
+    actors = [
+        "George Klooney",
```
Oleksl888, line 23: You could make it as a list of tuples and so you wouldn't need to use split() method later

Oleksl888 (approved):
Approved with a note 🥈

PR #956, opened 2024-08-11
Solution

wataschi:


settings.py
```diff
@@ -18,3 +18,6 @@
 USE_TZ = False
 
 INSTALLED_APPS = ("db",)
+
+ALLOWED_HOSTS = ["*"]
+DEBUG = "TRUE"
```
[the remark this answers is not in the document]
wataschi: I wanted to run the main.py file to check the code works and I get an error:
"CommandError: You must set settings.ALLOWED_HOSTS if DEBUG is False."
It seems that the problem was in the settings.py file, because after adding ALLOWED_HOSTS [*] everything worked correctly.
I suppose I should have removed these lines after testing?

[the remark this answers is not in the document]
Dimosphen1: Yes, `settings.py` should not be present in the PR (pull request), remove it, please

main.py
```diff
@@ -1,7 +1,44 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Actor, Genre
 
 
 def main() -> QuerySet:
-    pass
+
+    # Create
```
Dimosphen1, line 9: Remove these comments, there is no need for them. The code is self-explicit enough

main.py
```diff
@@ -1,7 +1,44 @@
...
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson"),
+    ]
+    for first_name, last_name in actors:
+        Actor.objects.create(first_name=first_name, last_name=last_name)
+
+    # Update
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+    (Actor.objects.filter(first_name="George", last_name="Klooney")
+     .update(last_name="Clooney"))
```
Dimosphen1, line 32: Instead of enveloping this expression inside of `()`, remove them and make line breaking after `(`, so that new line would not start with `.update(...`

Dimosphen1 (changes requested):

main.py
```diff
@@ -1,7 +1,43 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
```
pavlopro, line 3: Add blank line between different import groups

pavlopro (approved):
Good Job!

PR #959, opened 2024-08-14
Solution

DSahalatyi:


main.py
```diff
@@ -2,6 +2,55 @@
...
+    ("Scarlett", "Keegan"),
+    ("Will", "Smith"),
+    ("Jaden", "Smith"),
+    ("Scarlett", "Johansson")
+]
+
 
 def main() -> QuerySet:
-    pass
+    for genre in GENRES:
```
Oleksl888, line 20: Genres and Actors are used as global variables in your code which is not a good practice - you either declare them inside main or pass to main() as arguments

main.py
```diff
@@ -2,6 +2,55 @@
...
+        )
+
+    Genre.objects.filter(
+        name="Dramma",
+    ).update(name="Drama")
+
+    Actor.objects.filter(
+        first_name="George",
+        last_name="Klooney"
+    ).update(last_name="Clooney")
```
Oleksl888, line 36: Would make sense to break down main function is smaller specialized functions to structure your code better. E.g separate function for creating, updating, deleting..

Oleksl888 (changes requested):

main.py
```diff
@@ -1,7 +1,134 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from typing import Any
```
Oleksl888, line 4: must be a blank line between since typing is build in library and django is external package

Oleksl888 (approved):
It is a nice solution with parametrizing input, but I would say it is an overkill for a task like this. You could simply break down your functions by create/update/delete and do those actions on genres and actors in one function like in the below:
```
def delete() -> None:

Oleksl888:
> It is a nice solution with parametrizing input, but I would say it is an overkill for a task like this. You could simply break down your functions by create/update/delete and do those actions on genres and actors in one function like in the below:
> 
> ```
> def delete() -> None:
>     Genre.objects.filter(name="Action").delete()
>     Actor.objects.filter(first_name="Scarlett").delete()
> ```

Good job on your work anyway 🌟

PR #960, opened 2024-08-15
Solution

dartomOOv:


main.py
```diff
@@ -2,6 +2,55 @@
...
+    Actor.objects.filter(
+        first_name="Scarlett"
+    ).delete()
+
+    return Actor.objects.filter(
+        last_name="Smith"
+    ).order_by("first_name")
+
+
+genres = ["Western", "Action", "Dramma"]
```
Oleksl888, line 48: i guess this should be moved inside main function because now it is being passed as global

main.py
```diff
@@ -2,6 +2,55 @@
...
+        Genre.objects.create(
+            name=genre
+        )
+    for first_name, last_name in actors:
+        Actor.objects.create(
+            first_name=first_name.strip(),
+            last_name=last_name.strip()
+        )
+
+    Actor.objects.filter(
```
Oleksl888, line 19: here would make sense to break down main function is smaller specialized functions to structure your code better. E.g separate function for creating, updating, deleting..

Oleksl888 (changes requested):

main.py
```diff
@@ -2,6 +2,68 @@
 
 from django.db.models import QuerySet
 
+from db.models import Actor, Genre
+
 
 def main() -> QuerySet:
-    pass
+    def create(
```
Oleksl888, line 9: I didn't expect you to define these functions inside main(), it is possible, but it would make more sense to have them defined above main

Oleksl888 (approved):
Well done! But please check last comment

PR #969, opened 2024-08-21
Solution

EugeneDolinskiy:


main.py
```diff
@@ -1,7 +1,44 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
```
pavlopro, line 3: Add blank line between different import groups

pavlopro (changes requested):

main.py
pavlopro: not resolved

Dimosphen1: There is no need to add two blank lines, add one blank line between different groups of imports to follow PEP-8 in the future

Dimosphen1 (approved):
LGTM, one comment is left for consideration

PR #970, opened 2024-08-21
task done

Sergi0bbb:


main.py
```diff
@@ -1,7 +1,28 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Actor, Genre
```

main.py
```diff
@@ -1,7 +1,28 @@
...
 
 
 def main() -> QuerySet:
-    pass
+    Genre.objects.create(name="Western")
```
Oleksl888, line 8: To much repetition in the code below - read p2 adn p3 of checklist and adjust accordignly.

main.py
```diff
@@ -1,7 +1,28 @@
...
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+    Actor.objects.filter(last_name="Klooney").update(last_name="Clooney")
+    Actor.objects.filter(
+        first_name="Kianu",
+        last_name="Reaves"
+    ).update(first_name="Keanu", last_name="Reeves")
+
+    Genre.objects.filter(name="Action").delete()
+    Actor.objects.filter(first_name="Scarlett").delete()
```
Oleksl888, line 26: Also would make sense to break down main function is smaller specialized functions to structure your code better. E.g separate function for creating, updating, deleting..

Oleksl888 (changes requested):

PR #973, opened 2024-08-22
Solution Politov

Alexandr-Politov:


main.py
```diff
@@ -1,7 +1,38 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Actor, Genre
```

main.py
```diff
@@ -1,7 +1,38 @@
...
+
+    new_actors = ["George Klooney", "Kianu Reaves",
+                  "Scarlett Keegan", "Will Smith",
+                  "Jaden Smith", "Scarlett Johansson"]
+    for actor in new_actors:
+        first, last = actor.split(" ")
+        Actor.objects.create(first_name=first, last_name=last)
+
+    list_of_genres_updates = [("Dramma", "Drama")]
+    for old_genre, new_genre in list_of_genres_updates:
```
Oleksl888, line 21: no need for for loop here because you update only one item :)

main.py
```diff
@@ -1,7 +1,38 @@
...
 
 
 def main() -> QuerySet:
-    pass
+
+    new_genres = ["Western", "Action", "Dramma"]
+    for genre in new_genres:
+        Genre.objects.create(name=genre)
+
+    new_actors = ["George Klooney", "Kianu Reaves",
```
Oleksl888, line 13: It would be better to store a list of tuples where first element would be first name and second element last name. Then you will not need to call split() method and just use tuple unpacking which would be more pythonic` approach.
Also I would suggest to place each element of the list from new line

main.py
```diff
@@ -1,7 +1,38 @@
...
+    list_of_actors_updates = [("George Klooney", "George Clooney"),
+                              ("Kianu Reaves", "Keanu Reeves")]
+    for old_actor, new_actor in list_of_actors_updates:
+        old_first, old_last = old_actor.split(" ")
+        new_first, new_last = new_actor.split(" ")
+        Actor.objects.filter(
+            first_name=old_first, last_name=old_last
+        ).update(first_name=new_first, last_name=new_last)
+
+    Genre.objects.filter(name="Action").delete()
```
Oleksl888, line 33: Would make sense to break down main function is smaller specialized functions to structure your code better. E.g separate function for creating, updating, deleting..

Oleksl888 (changes requested):

Oleksl888 (changes requested):
nothing was changed

Alexandr-Politov:
> nothing was changed

Oh. I'm so sorry. I still have some problems with resolving merging branches changes :(

oliasymak24 (approved):

PR #974, opened 2024-08-22
first_try

yashkunn:


main.py
```diff
@@ -1,7 +1,32 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Genre, Actor
```

main.py
```diff
@@ -1,7 +1,32 @@
...
 
 
 def main() -> QuerySet:
-    pass
+    genres = [("Western",), ("Action",), ("Dramma",)]
```
Oleksl888, line 8: genres could be a list of strings or tuple of strings - make it more simple

main.py
```diff
@@ -1,7 +1,32 @@
...
+        ("Scarlett", "Johansson"),
+    ]
+
+    for genre_name, in genres:
+        Genre.objects.create(name=genre_name)
+
+    for first_name, last_name in actors:
+        Actor.objects.create(first_name=first_name, last_name=last_name)
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
```
Oleksl888, line 24: Would make sense to break down main function is smaller specialized functions to structure your code better. E.g separate function for creating, updating, deleting..

Oleksl888 (changes requested):

PR #976, opened 2024-08-22
Solution

panichqa:


main.py
```diff
@@ -1,7 +1,34 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Actor, Genre
```

main.py
```diff
@@ -1,7 +1,34 @@
...
+        ("George", "Klooney"),
+        ("Kianu", "Reaves"),
+        ("Scarlett", "Keegan"),
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson"),
+    ]
+
+    for genre_name in genres:
+        Genre.objects.get_or_create(name=genre_name)
```
Oleksl888, line 19: we don't really need to use get_or_create in this scenario

main.py
```diff
@@ -1,7 +1,34 @@
...
+    for actor_first_name, actor_second_name in actors:
+        Actor.objects.get_or_create(
+            first_name=actor_first_name,
+            last_name=actor_second_name
+        )
+    Genre.objects.filter(name="Dramma").update(name="Drama")
```
Oleksl888, line 25: Would make sense to break down main function is smaller specialized functions to structure your code better. E.g separate function for creating, updating, deleting..

Oleksl888 (changes requested):
Please change

Oleksl888 (approved):
Good job! It's nice that you parametrized input - that is a way to go. But in this task since we don't have a lot of data to work with you could combine create/update/delete functions to work with both actors and genres. Like so:
```
def delete() -> None:

PR #978, opened 2024-08-22
Solution

dpiuro:


main.py
```diff
@@ -1,7 +1,49 @@
 import init_django_orm  # noqa: F401
 
+
+from db.models import Genre, Actor
 from django.db.models import QuerySet
```
Oleksl888, line 5: must be a blank line between these import types

main.py
```diff
@@ -1,7 +1,49 @@
...
 
 
 def main() -> QuerySet:
-    pass
+    genres = [
+        ("Western",),
```
Oleksl888, line 10: You can simplify this by storing genres in list of strings or tuple of strings

main.py
```diff
@@ -1,7 +1,49 @@
...
+
+    Actor.objects.filter(
+        first_name="Kianu",
+        last_name="Reaves"
+    ).update(
+        first_name="Keanu",
+        last_name="Reeves"
+    )
+
+    Genre.objects.filter(name="Action").delete()
```
Oleksl888, line 44: Would make sense to break down main function is smaller specialized functions to structure your code better. E.g separate function for creating, updating, deleting..

Oleksl888 (changes requested):

main.py
```diff
@@ -1,7 +1,50 @@
...
+
+
 def main() -> QuerySet:
-    pass
+    create_genres()
+    create_actors()
+    update_entries()
+    delete_entries()
+
+    result = Actor.objects.filter(last_name="Smith").order_by("first_name")
```
Oleksl888, line 48: you could pass the expression in return directly - without creating a variable

Oleksl888 (approved):
Well done! :star:

PR #979, opened 2024-08-22
Resolve

ihorutkin:


main.py
```diff
@@ -1,7 +1,68 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Genre, Actor
```

main.py
```diff
@@ -1,7 +1,68 @@
...
 
 
 def main() -> QuerySet:
-    pass
+    Genre.objects.create(name="Western")
```
Oleksl888, line 8: Check p2-p3 of cheklist. Use for loops to call create methods within to avoid repetition

main.py
```diff
@@ -1,7 +1,68 @@
...
+    Actor.objects.filter(
+        first_name="Kianu"
+    ).update(
+        first_name="Keanu",
+        last_name="Reeves"
+    )
+    Genre.objects.filter(
+        name="Action"
+    ).delete()
+    Actor.objects.filter(
```
Oleksl888, line 56: Would make sense to break down main function is smaller specialized functions to structure your code better. E.g separate function for creating, updating, deleting..

main.py
```diff
@@ -1,7 +1,68 @@
...
+        first_name="Scarlett"
+    ).delete()
+    sorted_actors = Actor.objects.filter(
+        last_name="Smith"
+    ).order_by("first_name")
+    return sorted_actors
+
+
+if __name__ == "__main__":
```
Oleksl888, line 65: as per checklist this should not be included

Oleksl888 (changes requested):

main.py
```diff
@@ -2,6 +2,80 @@
...
+        Genre.objects.filter(
+            name="Action"
+        ).delete()
+    else:
+        Actor.objects.filter(
+            first_name="Scarlett"
+        ).delete()
+
+
+def sort() -> QuerySet:
```
Oleksl888, line 52: sort is not a very good name because it shadows build-in methods

main.py
```diff
@@ -2,6 +2,80 @@
 
 from django.db.models import QuerySet
 
+from db.models import Genre, Actor
+
+
+def create(data: list) -> None:
```
Oleksl888, line 8: Is ok, but a bit overcomplicated - you could just run two for loops - one for creating Genres and one for creating actors. But I see what you did and it's great that you parametrized input 👏

Oleksl888 (approved):
Well done!

PR #981, opened 2024-08-23
Solution

mnezbrytska:


main.py
```diff
@@ -2,6 +2,39 @@
...
+    actors = [
+        ("George", "Klooney"),
+        ("Kianu", "Reaves"),
+        ("Scarlett", "Keegan"),
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson")
+    ]
+    Actor.objects.bulk_create(
+        [Actor(first_name=fn, last_name=ln) for fn, ln in actors]
```
Oleksl888, line 22: use more descriptive names in unpacking - first_name - last_name would be great

Oleksl888 (changes requested):

main.py
```diff
@@ -1,7 +1,65 @@
-import init_django_orm  # noqa: F401
```
Oleksl888, line 1: this should not have been deleted

Oleksl888, line 1: this should not have been deleted

main.py
```diff
@@ -1,7 +1,65 @@
...
-
 from django.db.models import QuerySet
```
Oleksl888, line 1: and here should be a blank line because it's different imports

main.py
```diff
@@ -1,7 +1,65 @@
...
+    Actor.objects.filter(first_name="Scarlett").delete()
+
+
+def get_smith_actors() -> QuerySet:
+    return Actor.objects.filter(last_name="Smith").order_by("first_name")
 
 
 def main() -> QuerySet:
-    pass
+    create_genres()
```
Oleksl888, line 59: wouldn't make sense to make a function just to do one action - you could combine your functions by action type : create (will create both genres and actors), update(will update both) and same with delete

Oleksl888 (changes requested):

main.py
```diff
@@ -2,6 +2,56 @@
...
+        first_name="Keanu",
+        last_name="Reeves")
+
+
+def delete_actors_and_genres() -> None:
+    Genre.objects.filter(name="Action").delete()
+    Actor.objects.filter(first_name="Scarlett").delete()
+
+
+def get_() -> QuerySet:
```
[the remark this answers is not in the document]
mnezbrytska: Yes, of course, I didn't think of that

PR #984, opened 2024-08-23
solution for genres-and-actors

vladislav-tsybuliak1:


main.py
```diff
@@ -2,6 +2,49 @@
...
+            first_name=first_name,
+            last_name=last_name
+        )
+
+    Genre.objects.filter(
+        name="Dramma"
+    ).update(
+        name="Drama"
+    )
+    Actor.objects.filter(
```
Oleksl888, line 33: Would make sense to break down main function is smaller specialized functions to structure your code better. E.g separate function for creating, updating, deleting..

Oleksl888 (approved):
Good solution. But check the comment - you can make it even better 😉

PR #985, opened 2024-08-23
Solution

AndriiHamasa:


main.py
```diff
@@ -2,6 +2,46 @@
...
+
+    for genre in genres_to_create:
+        Genre.objects.create(name=genre)
+
+    for first_name, last_name in actors_to_create:
+        Actor.objects.create(first_name=first_name, last_name=last_name)
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+
+    actors_to_update = {
```
Oleksl888, line 27: that is actually quiet complicated code just to update two records, but ok let it be :)

main.py
```diff
@@ -2,6 +2,46 @@
...
+        old_first_name, old_last_name = old_name
+        new_first_name = new_name.get("first_name", old_first_name)
+        new_last_name = new_name.get("last_name", old_last_name)
+
+        Actor.objects.filter(
+            first_name=old_first_name,
+            last_name=old_last_name
+        ).update(
+            first_name=new_first_name, last_name=new_last_name
+        )
```
Oleksl888, line 42: You can break down main function in smaller specialized functions to structure your code better. E.g separate function for creating, updating, deleting..

Oleksl888 (approved):
Good job, but please check last comments

main.py
AndriiHamasa: Yeah, I though about it, thanks

main.py
AndriiHamasa: I thought I needed to adapt to both options at the same time

PR #988, opened 2024-08-24
Solution

olena4reunited:


main.py
```diff
@@ -2,6 +2,104 @@
...
+def create_genre(data: list[str]) -> None:
+    if isinstance(data[0], str):
+        for row in data:
+            Genre.objects.create(name=row)
+
+
+def create_actor(data: list[tuple[str, str]]) -> None:
+    if isinstance(data[0], tuple):
+        for row in data:
+            Actor.objects.create(first_name=row[0], last_name=row[1])
```
Oleksl888, line 17: better use unpacking instead of indexing

main.py
```diff
@@ -2,6 +2,104 @@
...
+) -> None:
+    if isinstance(data, str) and isinstance(new_data, str):
+        Genre.objects.filter(name=data).update(name=new_data)
+    if isinstance(data, list) and isinstance(new_data, list):
+        if isinstance(data[0], str):
+            for i in range(len(data)):
+                Genre.object.filter(name=data[i]).update(name=new_data[i])
+
+
+def update_actor(
```
Oleksl888, line 32: This is real nice what you did but here in this task you could be more simple since you have only few speicific names to update. You can break down your functions by create/update/delete and do thos actions on genres and actors in one function. Let's keep it simple like in an example:
```
def delete() -> None:

Oleksl888 (changes requested):

PR #991, opened 2024-08-25
Solution

Bogdan-Rylov:


settings.py
```diff
@@ -15,6 +15,8 @@
     }
 }
 
+ALLOWED_HOSTS = ["*"]
```
Oleksl888, line 18: please revert changes to this file

main.py
```diff
@@ -1,7 +1,49 @@
...
-
 from django.db.models import QuerySet
 
+import init_django_orm  # noqa: F401
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
-    pass
+    Genre.objects.bulk_create([
```
Oleksl888, line 8: follow the checklist p2 and p3: prepare data in tuples and then create entries to db in a cycle

main.py
```diff
@@ -1,7 +1,49 @@
...
+    ).update(last_name="Clooney")
+    Actor.objects.filter(
+        first_name="Kianu",
+        last_name="Reaves"
+    ).update(
+        first_name="Keanu",
+        last_name="Reeves"
+    )
+
+    Genre.objects.filter(
```
Oleksl888, line 38: Try to break down main function is smaller specialized functions to structure your code better. E.g separate function for creating, updating, deleting (no need to parametrize arguments, it will be enough to keep expressions as they are right now, just group them under different functions)

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #995, opened 2024-08-27
Solution

NazarSerdiuk1:


main.py
```diff
@@ -1,7 +1,41 @@
 import init_django_orm  # noqa: F401
```
pavlopro, line 2: There is no need blank lines between those two imports

main.py
```diff
@@ -1,7 +1,41 @@
...
 
+
 from django.db.models import QuerySet
 
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
-    pass
+    genres = [
```
pavlopro, line 10: Why do you need those tuples in list?
simplify this

pavlopro (changes requested):

main.py
```diff
@@ -1,7 +1,34 @@
 import init_django_orm  # noqa: F401
-
 from django.db.models import QuerySet
```
Dimosphen1, line 2: Add one blank line between different import groups to follow PEP-8 in the future

Dimosphen1 (approved):
LGTM, one comment is left for consideration

PR #1014, opened 2024-09-17
Solution?

Hennadii29:
why my tables not updating with provided data?

### DIFF
--- a/db/models.py
+++ b/db/models.py
@@ -1,1 +1,10 @@
 from django.db import models
+
+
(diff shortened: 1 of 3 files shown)


vasylhnatiuk (changes requested):
add migrations

Hennadii29:
> add migrations

i did makemigrations and migrate in terminal, now tables are not empty, but still not able to succesfully push it...

PR #1022, opened 2024-09-19
solution 01

AversArthur:


AversArthur:
sorry, I getting this error, and I don't find a way to fix it 
ERROR tests/test_main.py - RuntimeError: Database access not allowed, use the "django_db" mark, or the "db" or "transactional_db" fixtures to enable it.

AversArthur:
Screenshot 2024-09-20 at 11 47 04
but it seems like I did.

main.py
```diff
@@ -2,6 +2,30 @@
 
 from django.db.models import QuerySet
 
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
-    pass
+    Genre.objects.create(name="Western")
```
vasylhnatiuk, line 9: use loop

main.py
```diff
@@ -2,6 +2,30 @@
...
+    Genre.objects.create(name="Action")
+    Genre.objects.create(name="Dramma")
+
+    Actor.objects.create(first_name="George", last_name="Klooney")
```
vasylhnatiuk, line 13: use loop

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
fix tests, add migrations

AversArthur:
I do not get it. When I run pytest on local, there is no issues, only a couple warnings. But when I push it, it not able to pass tests.

LLkaia (changes requested):
Your PR changes:
image

You added two new models. If so, then these changes should lead to creation of new tables. That's why we should see new migrations in your PR. Check screenshot - there is no such

Mb you have them on your local, but have not pushed them to remote branch. Please, do this.

AversArthur:
> Your PR changes: image
> 
> You added two new models. If so, then these changes should lead to creation of new tables. That's why we should see new migrations in your PR. Check screenshot - there is no such
> 
> Mb you have them on your local, but have not pushed them to remote branch. Please, do this.

Hey mate! you were right, thanks! 
I notice my migrate files wasn't choose automatically to PR. I have to select them manually for some reason

vasylhnatiuk (approved):

PR #1042, opened 2024-10-19
Created solution with actors and genres

Mxbely:


main.py
```diff
@@ -1,7 +1,45 @@
...
+        ("Scarlett", "Keegan"),
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson"),
+    ]
+    for genre in genres:
+        Genre.objects.create(name=genre)
+
+    for first, last in actors:
+        Actor.objects.create(first_name=first, last_name=last)
```
Arsen-hrynevych, line 21: get_or_create instead of create: Using get_or_create ensures that you don’t create duplicates if the records already exist. It also avoids raising an error if you attempt to create a record that already exists.

Arsen-hrynevych (approved):
ok

main.py
Mxbely: okey, but i already saw, that it's the next topic

PR #1050, opened 2024-10-26
solution

dmytrik:


VladyslavBon (changes requested):
You have to comit your migration file to pass tests on github

PR #1063, opened 2024-11-01
Solution

Le0n-K:


main.py
```diff
@@ -1,7 +1,48 @@
...
+from db.models import Genre, Actor
 
 
 def main() -> QuerySet:
-    pass
+    genres = ["Western", "Action", "Dramma"]
+    actors_name = ["George", "Kianu", "Scarlett",
+                   "Will", "Jaden", "Scarlett"]
+    actors_surname = ["Klooney", "Reaves", "Keegan",
+                      "Smith", "Smith", "Johansson"]
```
YaroslavBordovoy, line 12: I recommend you create one variable "actors" with a list of tuples [("George", "Klooney"), ...]. And after that in for loop use this variable. Also its recommendation from the checklist

YaroslavBordovoy (changes requested):

main.py
Le0n-K: ThumbsUpSimonCowellGIF

PR #1065, opened 2024-11-01
solution

serhiik05:
I Know not best practise , found this bulk_create method before read Checklist.md
Look not that bad


skyfoxwork (approved):
Good job!
Interesting method bulk_create!
I don't know it is good or bad, but looks like not bad )
I think you will write less code if you will use loops.

Arsen-hrynevych (approved):
ok

PR #1067, opened 2024-11-01
'solution'

spa7id:


dxrrkwm (changes requested):
change main.py according to checklist and you can also make it using only main function :)

dxrrkwm (commented):
use for loop to create actors and genres according to checklist

PR #1070, opened 2024-11-03
Solution

MatvijSafonov:


main.py
```diff
@@ -1,7 +1,42 @@
...
+from db.models import Genre, Actor
 
 
 def main() -> QuerySet:
-    pass
+    genres = [
+        ("Western",),
+        ("Action",),
+        ("Dramma",)
+    ]
```
94nj111, line 12: It seems that tuples are not needed here

94nj111 (changes requested):

PR #1077, opened 2024-11-04
Solution

FursykIvan:


sumseer (commented):
Looks like you need to add migrations to your PR. You can do this by running 'git add db/migrations' and then creating a new commit and pushing it to the PR.

sumseer (approved):
Looks good now!

Den-k0 (approved):

PR #1135, opened 2025-01-05
solution

DanilTrukhanov:


taraskhom (approved):
You have a cool implementation of the for loop to create objects, I completely forgot about it

PR #1137, opened 2025-01-05
Solution

taraskhom:


main.py
```diff
@@ -1,7 +1,43 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Actor, Genre
 
 
 def main() -> QuerySet:
-    pass
+    Genre.objects.bulk_create(
```
ruslanyemets, line 8: You can use the list of genres and the 'for' loop in order not to repeat yourself

main.py
```diff
@@ -1,7 +1,43 @@
...
+        [
+            Genre(name="Western"),
+            Genre(name="Action"),
+            Genre(name="Dramma")
+        ]
+    )
+    Actor.objects.bulk_create(
```
ruslanyemets, line 15: You can put values in the list of tuples for iterating by actors and use the 'for' loop

ruslanyemets (approved):
Good job! But you can make some improvements to make your code more perfect

main.py
L1nk3rrr: Agree with this comment

main.py
L1nk3rrr: Agree with this comment

L1nk3rrr (changes requested):
Well done! Please review comment from Ruslan. They are related to the checklist that you have for a task. Try the DRY!
DRY - don't repeat yourself.

main.py
```diff
@@ -1,7 +1,56 @@
...
+        name="Action"
+    ).delete()
+
+    Actor.objects.filter(
+        first_name="Scarlett"
+    ).delete()
+
+    return Actor.objects.filter(
+        last_name="Smith"
+    ).order_by("first_name")
```
petrykivd, line 56: If the methods can be written in one line, it is better not to split it into new lines.
Check it and fix in another places

petrykivd (approved):
GJ! 👍

PR #1144, opened 2025-01-06
Solution

bogdAAAn1:


4ikkamony (approved):
Nice work, you just need to fix flake8 errors, replace single quotes with double ones

PR #1152, opened 2025-01-07
Solution

MindDevastation:


Professor-Douel (changes requested):
add the initial.py to pullrequest

PR #1165, opened 2025-01-09
Solution

shretsker:


main.py
```diff
@@ -4,4 +4,24 @@
 
 
 def main() -> QuerySet:
-    pass
+    Genre.objects.create(name="Western")
```
yelizaveta-zh, line 7: Use here "for" loop for create Genre objects, and the same with Actor objects (this is recommendation from checklist).
Example of structure:
`for item_first, item_second in items:
       Model.objects.create(
           first=item_first,
           second=item_second
       )`

yelizaveta-zh (changes requested):
Your code is good, but read the checklist and add loops to avoid repeating a lot of code.
Also flake8 don't pass, maybe you can fix it if add
`from db.models import Genre, Actor`
after imports already added in the file

main.py
```diff
@@ -2,6 +2,44 @@
 
 from django.db.models import QuerySet
 
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
-    pass
+
+    genres = [
```
Nattalli, line 10: why don't you use list?

Nattalli (changes requested):

PR #1166, opened 2025-01-09
solution_v.1.0

yelizaveta-zh:


main.py
```diff
@@ -2,6 +2,48 @@
...
 def main() -> QuerySet:
-    pass
+
+    genres = [
+        ("Western",),
+        ("Action",),
+        ("Dramma",)
+    ]
+    for name, in genres:
+        Genre.objects.create(name=name)
```
petrykivd, line 16: I think you can simplify it to list of strings 😄

main.py
```diff
@@ -2,6 +2,48 @@
...
+
+    Genre.objects.filter(name="Action").delete()
+
+    Actor.objects.filter(first_name="Scarlett").delete()
+
+    actors_smith = Actor.objects.filter(
+        last_name="Smith"
+    ).order_by("first_name")
+
+    return actors_smith
```
petrykivd, line 49: You can avoid unnecessary variables and return the filtered actor at once

petrykivd (changes requested):

PR #1253, opened 2025-03-18
Solution

Oneyura:


Nikita-pyth (approved):
Nice that you came up with using loops, i wasn't that smart haha

PR #1261, opened 2025-03-19
solution (but cgi issue on my PC)

cemen2009:


AndyStarGitHub (approved):
Good job. For the future - it is possible to create the list of records in the loop, or use bulkcreate.

PR #1419, opened 2025-07-24
Solution.

Otst0y:


Orixara (approved):
great job, I'm thinking on the same way

PR #1433, opened 2025-07-26
create migration and QuerySet

SerhiiBasok:


Artcrafterrra (approved):
Great solution with .bulk_create. Well done!

PR #1435, opened 2025-07-26
Solution

Artcrafterrra:


main.py
```diff
@@ -1,7 +1,31 @@
...
+        Genre(name="Dramma"),
+    ])
+    Actor.objects.bulk_create([
+        Actor(first_name="George", last_name="Klooney"),
+        Actor(first_name="Kianu", last_name="Reaves"),
+        Actor(first_name="Scarlett", last_name="Keegan"),
+        Actor(first_name="Will", last_name="Smith"),
+        Actor(first_name="Jaden", last_name="Smith"),
+        Actor(first_name="Scarlett", last_name="Johansson"),
+    ])
```
Kostenikov, line 20: As checklist says: `Use the for loop in order not to repeat yourself`.
Create a list with Genre strings, and list with tuples with Actors for iteration in a loop

Kostenikov (changes requested):

main.py
Artcrafterrra: @Kostenikov You're right in the case where I would create each object individually using .create — then I would have to duplicate a line like Genre.objects.create(Actor(first_name="George", last_name="Klooney")), Genre.objects.create(Actor(first_name="Kianu", last_name="Reaves")), and so on for each request, since .create does not accept more than one argument.

However, I intentionally used .bulk_create — it's a different command that allows sending all the data in a single query, which is faster than executing each query separately, even if done in a loop. That would still result in multiple queries. On the other hand, .bulk_create allows for a single query, which makes the program run faster — that's why I used it.

Based on the above, it’s clear that there is no DRY violation here.

Kostenikov: After some research I agree with you, it might be faster, but I thought a loop was needed in this task. 
Well done!🎉

Kostenikov (approved):

PR #1438, opened 2025-07-27
solution py genres and actors

PavelSemenikhin:


main.py
```diff
@@ -1,7 +1,33 @@
...
+
+    for first_name, last_name in list_of_actors:
+        Actor.objects.create(first_name=first_name, last_name=last_name)
+
+    Genre.objects.filter(name="Dramma").update(name="Drama")
+    Actor.objects.filter(last_name="Klooney").update(last_name="Clooney")
+    Actor.objects.filter(
+        first_name="Kianu",
+        last_name="Reaves"
+    ).update(first_name="Keanu", last_name="Reeves")
```
dimashevchukk, line 29: using same formatting would be more readable

main.py
```diff
@@ -1,7 +1,33 @@
...
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson")
+    ]
+
+    for genre in list_of_genres:
+        Genre.objects.create(name=genre)
+
+    for first_name, last_name in list_of_actors:
+        Actor.objects.create(first_name=first_name, last_name=last_name)
```
dimashevchukk, line 22: u can also use bulk_create to do one sql query instead of loop

dimashevchukk (approved):

dimashevchukk (approved):

PR #1443, opened 2025-07-27
Solution

DenisPrudius:


main.py
```diff
@@ -2,6 +2,75 @@
 
 from django.db.models import QuerySet
 
+from db.models import Genre, Actor
+
 
 def main() -> QuerySet:
-    pass
+    Genre.objects.create(
```
MorsImmortalis05, line 9: it's more convenient to use bulk_create here, like
`Genre.objects.bulk_create(
    [
        Genre(name="Action),
        Genre(name=...)
    ]`
and same goes for Actors

MorsImmortalis05 (approved):

PR #1445, opened 2025-07-27
Solution

Plovuezz:


bohdankolisnyk04 (approved):
good job! but bro.... i dont undestand your second part, now its so hard for me
if u can and want - write me! I want now how its work....

PR #1580, opened 2025-11-26
Solution

VYGAKOOL:


Iventyk (approved):
Great Job!

I had the same error on this task. You have to "Git add" the migrations files to push them on git
Use the command "git status" to check untracked files, then do "Git add file_name" and only after that commit and push on git

PR #1581, opened 2025-11-26
Solution

Ann-Riabykina:


tradewcs (approved):
you should commit your db migrations to pass tests

PR #1729, opened 2026-04-01
Solution

dmdva:


main.py
```diff
@@ -2,6 +2,75 @@
...
+    actresses = [
+        Actor(
+            first_name="Scarlett",
+            last_name="Keegan"
+        ),
+        Actor(
+            first_name="Scarlett",
+            last_name="Johansson"
+        ),
+    ]
```
vsmutok: Write it the same way as in `genres`.
There is no need to separate `actresses` and `actors`.

vsmutok (changes requested):
Great!

Boyok1337 (approved):
LGTM!

PR #1774, opened 2026-04-29
Solution

herman-py:


SerhiiHryniuk (approved):

main.py
```diff
@@ -1,7 +1,34 @@
 import init_django_orm  # noqa: F401
-
 from django.db.models import QuerySet
+from db.models import Genre, Actor
 
 
 def main() -> QuerySet:
-    pass
+    genres = [("Western",), ("Action",), ("Dramma",)]
```
VadimPusiak788, line 7: using a simple list of strings and bulk_create

main.py
```diff
@@ -1,7 +1,34 @@
...
+        ("Scarlett", "Keegan"),
+        ("Will", "Smith"),
+        ("Jaden", "Smith"),
+        ("Scarlett", "Johansson"),
+    ]
+
+    for (name,) in genres:
+        Genre.objects.create(name=name)
+
+    for first_name, last_name in actors:
```
VadimPusiak788, line 20: Use bulk_create to insert records in a single batch query instead of N separate queries.

VadimPusiak788 (changes requested):

PR #1788, opened 2026-05-03
create table Genre and create table Actor

ilonanikolaychuk052-afk:


main.py
```diff
@@ -1,7 +1,34 @@
...
+    Genre.objects.create(name="Western")
+    genre_action = Genre.objects.create(name="Action")
+    genre_dramma = Genre.objects.create(name="Dramma")
+    actor_george = Actor.objects.create(
+        first_name="George",
+        last_name="Klooney"
+    )
+    actor_kianu = Actor.objects.create(
+        first_name="Kianu",
+        last_name="Reaves"
```
petrykivd: Consider grouping genres into a list and creating them in a loop instead of calling create() separately for each one.

main.py
```diff
@@ -1,7 +1,34 @@
...
+    )
+    Actor.objects.create(first_name="Scarlett", last_name="Keegan")
+    Actor.objects.create(first_name="Will", last_name="Smith")
+    Actor.objects.create(first_name="Jaden", last_name="Smith")
+    Actor.objects.create(first_name="Scarlett", last_name="Johansson")
+
```
petrykivd: Same here — a list of tuples + a loop would be cleaner and easier to extend.

main.py
```diff
@@ -1,7 +1,34 @@
...
+    genre_dramma.name = "Drama"
+    genre_dramma.save()
+    actor_george.last_name = "Clooney"
+    actor_george.save()
+    actor_kianu.first_name = "Keanu"
+    actor_kianu.last_name = "Reeves"
+    actor_kianu.save()
```
petrykivd: Prefer .filter().update() over fetching an object and calling .save() — it's one DB query instead of two. Same applies to actor updates below.

petrykivd (changes requested):

VadimPusiak788 (approved):

PR #1798, opened 2026-05-11
'Solution'

Dikikim:


main.py
```diff
@@ -1,7 +1,30 @@
 import init_django_orm  # noqa: F401
 
 from django.db.models import QuerySet
+from db.models import Genre, Actor
 
 
 def main() -> QuerySet:
-    pass
+    Genre.objects.create(name="Western")
```
VadimPusiak788: Create lists of genre names and actor data, then build Genre and Actor instances from those lists. Use bulk_create() to reduce the number of database queries when inserting multiple objects.

VadimPusiak788 (changes requested):

VadimPusiak788 (approved):

